PR #196, opened 2023-04-04
[ADD] Background tag filtering UI

zorgjbe:
This makes the tag filter UI work using the new tags. It works, though it only filters one tag at the moment and the UI is pretty jank, so it's definitely a work-in-progress.

### DIFF
--- a/pandora-client-web/src/components/chatroomAdmin/chatroomAdmin.scss
+++ b/pandora-client-web/src/components/chatroomAdmin/chatroomAdmin.scss
@@ -131,7 +131,7 @@
 			display: none;
 			position: absolute;
 			max-height: 50vh;
-            overflow-y: auto;
+			overflow-y: auto;
 			background-color: $grey-lighter;
 			box-shadow: 0.3em 0.3em 0.2em $drop-shadow;
 			padding: 0.8em 1em;
--- a/pandora-client-web/src/components/chatroomAdmin/chatroomAdmin.scss
+++ b/pandora-client-web/src/components/chatroomAdmin/chatroomAdmin.scss
@@ -183,3 +183,10 @@
 		max-width: 100%;
 	}
 }
+
+.backgroundInfo {
+	img {
+		max-width: 46em;
+		max-height: 46em;
+	}
+}
--- a/pandora-client-web/src/components/chatroomAdmin/chatroomAdmin.scss
+++ b/pandora-client-web/src/components/chatroomAdmin/chatroomAdmin.scss
@@ -55,7 +55,6 @@
 }
 
 .backgroundSelect {
-
 	max-width: 95vw;
 	max-height: 60vh;
 	display: flex;
@@ -80,7 +79,7 @@
 		}
 
 		.header-tags {
-			@include flex(row);
+			@include flex(row, wrap);
 			gap: 0.5em;
 			width: 100%;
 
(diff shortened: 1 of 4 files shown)


Sekkmer:
image

pandora-client-web/src/components/chatroomAdmin/chatroomAdmin.tsx
```diff
@@ -507,18 +534,18 @@ function BackgroundSelectDialog({ hide, current, select }: {
 
 	const availableBackgrounds = useMemo(() => assetManager.getBackgrounds(), [assetManager]);
 	const [nameFilter, setNameFilter] = useState('');
-	/*
-	 * TODO: Add a tag based filter to the dialog in a later version
-	 * const [tagFilter, setTagFilter] = useState('');
-	 */
+	const [tags, setTags] = useState<readonly string[]>([...assetManager.backgroundTags.keys()]);
+	const knownTagCategories = useMemo(() => uniq([...assetManager.backgroundTags.values()].map((tag) => tag.category)), [assetManager.backgroundTags]);
```
Jomshir98, line 538: Selection behavior:
Having all selected by default and unselecting feels very weird - one has to first unselect before selecting those that are wanted.
We think much better would be to have by default all unselected and have a special case that none selected is the same as all selected (empty filter is ignored)

Jomshir98, line 538: Combining logic:
Having the tags combine with "OR" logic across categories is weird, especially with the fact that everything is selected by default - this basically means that one has to first deselect everything in all categories and then select the one thing one is searching for.
I think the categories should combine with an "AND" - categories where nothing is selected (which would be default) are ignored, items within category are combined with "OR" and categories themselves are combined with "AND".
Example selection:
- Time: Night
- Location: Bar, Beach
- Ambience, Item, Space: [nothing]

This selects all backgrounds that have "Night" tag _and_ have at least one of ["Bar", "Beach"] tags.

pandora-client-web/src/components/chatroomAdmin/chatroomAdmin.scss
```diff
@@ -63,11 +63,36 @@
 	gap: 0.5em;
```
Jomshir98, line 63: I think the `.backgroundSelect` needs its `max-width` and `max-height` changed to be constant, not max (`width`&`height`), as it feels really clunky when everything moves under your hand when modifying filters.
We think the following values look good on both normal screens and small ones (phones):
```css
width: min(50em, 90vw);
height: min(40em, 90vh);
```

pandora-client-web/src/components/chatroomAdmin/chatroomAdmin.scss
```diff
@@ -63,11 +63,36 @@
...
+			column-gap: 0.6em;
+			width: 100%;
+
+			span {
+				margin: auto 0;
+			}
+		}
+
+		.header-tags {
+			@include flex(row);
```
Jomshir98, line 83: This needs to wrap for narrow screens.
```suggestion
			@include flex(row, wrap);
```

pandora-client-web/src/components/chatroomAdmin/chatroomAdmin.scss
```diff
@@ -101,6 +128,7 @@
 		}
 
 		.dropdown-content {
+			@include flex(column);
 			display: none;
 			position: absolute;
```
Jomshir98, line 133: This element can easily overflow the screen for long lists. The suggested solution isn't perfect, but solves most of the overflowing:
```suggestion
			position: absolute;
			max-height: 50vh;
			overflow-y: auto;
```

Jomshir98 (changes requested):

pandora-client-web/src/components/chatroomAdmin/chatroomAdmin.scss
Sekkmer: I didn't quite like the result for the height so I moved the dialog to the top so the selection doesn't jump around

Jomshir98 (approved):
Looks good to me and feels great to use!

PR #296, opened 2023-06-15
[ADD] AWS SES email transport handling

Sekkmer:
Our current free SMTP provider become unreliable so we need to move to a new service.

I put on-hold on this PR until AWS approved my production access request and I successfully use the service

### DIFF
--- a/pandora-server-directory/package.json
+++ b/pandora-server-directory/package.json
@@ -37,6 +37,8 @@
 		]
 	},
 	"dependencies": {
+		"@aws-sdk/client-ses": "^3.352.0",
+		"@aws-sdk/credential-provider-node": "^3.352.0",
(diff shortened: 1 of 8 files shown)


Jomshir98 (approved):
Looks good!
Little thing I stumbled upon: Do we actually need the AWS packages? It seems it supports sending using SMTP too: https://docs.aws.amazon.com/ses/latest/dg/send-email-smtp.html

Sekkmer:
> Looks good! Little thing I stumbled upon: Do we actually need the AWS packages? It seems it supports sending using SMTP too: https://docs.aws.amazon.com/ses/latest/dg/send-email-smtp.html

Yep, I know. I like the security key a bit more than the user and password and don't feel that AWS packages would hurt, I made it so if you run mock or smtp they are not even loaded

PR #297, opened 2023-06-16
[CHANGE] unify leave actions to the top right button

Sekkmer:
image

### DIFF
--- a/pandora-client-web/src/components/header/leaveButton.scss
+++ b/pandora-client-web/src/components/header/leaveButton.scss
@@ -4,6 +4,7 @@
 		flex-flow: column;
 		margin-bottom: 0.5em;
 		width: 30em;
+		user-select: text;
 
 		button {
 			margin-top: 1em;
--- a/pandora-client-web/src/components/header/leaveButton.scss
+++ b/pandora-client-web/src/components/header/leaveButton.scss
@@ -1,0 +1,14 @@
+.LeaveDialog {
+	fieldset {
+		display: flex;
+		flex-flow: column;
+		margin-bottom: 0.5em;
+		width: 30em;
+
+		button {
+			margin-top: 1em;
+			margin-left: auto;
+			width: 12em;
+		}
+	}
+}
--- a/pandora-client-web/src/components/header/leaveButton.scss
+++ b/pandora-client-web/src/components/header/leaveButton.scss
@@ -10,6 +10,7 @@
 			margin-top: 1em;
 			margin-left: auto;
 			width: 12em;
+			user-select: none;
 		}
 	}
 }
--- a/pandora-client-web/src/components/header/leaveButton.tsx
+++ b/pandora-client-web/src/components/header/leaveButton.tsx
@@ -1,4 +1,4 @@
-import React, { ForwardedRef, ReactElement, ReactNode, createContext, forwardRef, useCallback, useContext, useState } from 'react';
+import React, { ReactElement, createContext, useCallback, useContext, useState } from 'react';
 import logoutIcon from '../../assets/icons/logout.svg';
 import { HeaderButton } from './HeaderButton';
 import { useShardConnectionInfo } from '../gameContext/shardConnectorContextProvider';
@@ -12,7 +12,7 @@ import { PlayerCharacter } from '../../character/player';
 import { toast } from 'react-toastify';
 import { TOAST_OPTIONS_ERROR } from '../../persistentToast';
 import { ModalDialog } from '../dialog/dialog';
-import { Column, Row } from '../common/container/container';
+import { Column } from '../common/container/container';
 import './leaveButton.scss';
 import { HoverElement } from '../hoverElement/hoverElement';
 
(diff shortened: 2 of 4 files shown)


ClaudiaMia (commented):
I tested it!~

Works well and I like it, but I do have general thoughs that I would not see as change requests but more as food for thought:

- Do we really need the "Are you sure?"-step with another click for these buttons? I mean, we are already in a special dialog which prevents accidental triggering by its very nature.
- Maybe it would not fit visually to the rounded edges and counter the desire to reuse existing components, but it may be nice if this dialog opens snuggly fit in the very top right corner (or under the top menu bar aligned with the right edge) so you save on travel time to the buttons.

ClaudiaMia (changes requested):
Tested the latest version!~

When you click the leave room or change character button, the dialog remains open on the new screen.
It should automatically close after the button was pressed.

pandora-client-web/src/components/header/leaveButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+	);
+}
+
+function ChatRoomLeave(): ReactElement | null {
+	const player = usePlayer();
+	const room = useChatRoomInfo();
+
+	if (!player || !room) {
+		return null;
+	}
```
Jomshir98, line 56: I think instead of not showing it, it is much better to show the area, but instead of button show that you are not in a room (which will later on turn into that you are in a room that you cannot leave - your personal room)

pandora-client-web/src/components/header/leaveButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+	const characterData = usePlayerData();
+	const characterName = (characterData && !characterData.inCreation) ? characterData.name : null;
+
+	const onClick = useCallback((e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
+		e.preventDefault();
+		e.stopPropagation();
+		directoryConnector.sendMessage('disconnectCharacter', EMPTY);
+	}, [directoryConnector]);
+
+	if (!connectionInfo) return null;
```
Jomshir98, line 113: I think instead of not showing it, it is much better to show the area, but instead of button show that you don't have a character selected.

pandora-client-web/src/components/header/leaveButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+			<HeaderButton icon={ logoutIcon } iconAlt='Leave' onClick={ () => setShow(true) } title='Leave' />
+			{
+				show ? <DialogLeave /> : null
+			}
+		</leaveButtonContext.Provider>
+	);
+}
+
+function DialogLeave(): ReactElement {
+	const closeDialog = useContext(leaveButtonContext);
```
Jomshir98, line 36: Nit: Wouldn't it be much simpler to just pass it as a prop instead of context?

pandora-client-web/src/components/header/leaveButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+	const onRoomLeave = useCallback(() => {
+		directoryConnector.awaitResponse('chatRoomLeave', EMPTY)
+			.then((result) => {
+				if (result.result !== 'ok') {
+					toast(`Failed to leave room:\n${result.result}`, TOAST_OPTIONS_ERROR);
+				}
+			})
+			.catch((err) => {
+				GetLogger('LeaveRoom').warning('Error during room leave', err);
+				toast(`Error during room creation:\n${err instanceof Error ? err.message : String(err)}`, TOAST_OPTIONS_ERROR);
```
Jomshir98, line 78: Likely already was that way before, but should be fixed anyway:
```suggestion
				toast(`Error while leaving room:\n${err instanceof Error ? err.message : String(err)}`, TOAST_OPTIONS_ERROR);
```

pandora-client-web/src/components/header/leaveButton.scss
```diff
@@ -0,0 +1,14 @@
+.LeaveDialog {
+	fieldset {
```
Jomshir98, line 2: As you already show so much info, maybe it might be good to also make it copy-able using `user-select: text;`?

pandora-client-web/src/components/header/leaveButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+			<legend>Chat Room</legend>
+			<span>Name: { room.name }</span>
+			<span>Id: { room.id }</span>
+			{
+				!canLeave ? (
+					<HoverElement parent={ leaveButtonRef } className='action-warning'>
+						An item is preventing you from leaving the room.
+					</HoverElement>
+				) : null
+			}
```
Jomshir98, line 93: Idea (optional): Before there wasn't really a good place to put it, but now that there is lots of space, maybe it would be better to show this always instead of only on hover/hold?

Jomshir98 (changes requested):
Originally only wanted to leave two comments, but it turned into a full review~ :sweat_smile:

pandora-client-web/src/components/header/leaveButton.tsx
Sekkmer: I didn't want to pass that action all the way down so I felt a context result in a simpler code

pandora-client-web/src/components/header/leaveButton.tsx
Sekkmer: I was thinking about it but had no idea how to show it visually, so I just went with the copy >.<

Jomshir98 (approved):
Looks good to me!

ClaudiaMia (approved):
Got permission to approve now!~

PR #300, opened 2023-06-16
[FEATURE] attributes as atomic conditions

### LINKED ISSUE #268: Add image override also based on attributes
Currently, image overrides are based for example on modules or bone states, but they should also be possible to be triggered based on the attributes of items worn by the character.

Sekkmer:
allow an attribute to be required or not allowed if it is prefixed with `!`

resolves #268

### DIFF
--- a/pandora-client-web/src/editor/parsing.ts
+++ b/pandora-client-web/src/editor/parsing.ts
@@ -35,15 +35,15 @@ export function SerializeAtomicCondition(condition: Immutable<AtomicCondition>):
 		return `hand_${condition.armType}_${condition.side}${condition.operator}${condition.value}`;
 	} else if ('attribute' in condition) {
 		Assert(condition.attribute != null);
-		return `attribute ${condition.attribute}`;
+		return `a_${condition.attribute}`;
 	} else {
 		AssertNever(condition);
 	}
 }
 
 function ParseAtomicCondition(input: string, validBones: string[]): AtomicCondition {
-	if (input.startsWith('attribute')) {
-		const attribute = /^attribute\s+(!?[-_a-z0-9]+)\s*$/i.exec(input);
+	if (input.match(/^\s*a_/)) {
+		const attribute = /^\s*a_(!?[-_a-z0-9]+)\s*$/.exec(input);
 		if (!attribute) {
 			throw new Error(`Failed to parse attribute condition '${input}'`);
 		}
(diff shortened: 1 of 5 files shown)


pandora-client-web/src/editor/parsing.ts
```diff
@@ -24,19 +24,33 @@ function ParseFloat(input: string): number {
...
-
-	if ('armType' in condition && condition.armType != null)
+	} else if ('armType' in condition) {
+		Assert(condition.armType != null);
 		return `hand_${condition.armType}_${condition.side}${condition.operator}${condition.value}`;
-
-	AssertNever();
+	} else if ('attribute' in condition) {
+		Assert(condition.attribute != null);
+		return `attribute ${condition.attribute}`;
```
Jomshir98, line 38: This serialization won't work, as during deserialization, things are split by space first, as space splits condition and image... I really need to improve this :sweat_smile: 
```suggestion
		return `a_${condition.attribute}`;
```

pandora-client-web/src/editor/parsing.ts
```diff
@@ -24,19 +24,33 @@ function ParseFloat(input: string): number {
...
+	} else {
+		AssertNever(condition);
+	}
 }
 
 function ParseAtomicCondition(input: string, validBones: string[]): AtomicCondition {
+	if (input.startsWith('attribute')) {
+		const attribute = /^attribute\s+(!?[-_a-z0-9]+)\s*$/i.exec(input);
```
Jomshir98, line 46: And the deserialization will need to be updated too.

Jomshir98 (changes requested):

pandora-client-web/src/editor/parsing.ts
```diff
@@ -24,19 +24,33 @@ function ParseFloat(input: string): number {
...
+		Assert(condition.attribute != null);
+		return `a_${condition.attribute}`;
+	} else {
+		AssertNever(condition);
+	}
 }
 
 function ParseAtomicCondition(input: string, validBones: string[]): AtomicCondition {
+	if (input.match(/^\s*a_/)) {
+		const attribute = /^\s*a_(!?[-_a-z0-9]+)\s*$/.exec(input);
```
Jomshir98, line 46: Why the space? There can never be a space in here. Also case insensitive (or allowing A-Z) is needed, as literally all our attributes have capitalized names.

PR #301, opened 2023-06-17
[CHANGE] Chatroom rework

Jomshir98:
This PR refactors chatroom, splitting the code more, reusing more common components and cleaning up the looks a bit.
It also introduces new feature: Ability to configure ratio of graphics to chat, defaulting to 7:3 (4:6 for portrait).
Also disables the test popup that has no value and somehow made it in through the review, I will yet look for how it got in without being rejected.

### DIFF
--- a/pandora-client-web/src/common/useMediaQuery.ts
+++ b/pandora-client-web/src/common/useMediaQuery.ts
@@ -1,0 +1,20 @@
+import { useCallback, useDebugValue, useMemo, useSyncExternalStore } from 'react';
+
+export function useMediaQuery(query: string): boolean {
+	const match = useMemo(() => window.matchMedia(query), [query]);
+
+	const subscribe = useCallback((onChange: () => void) => {
+		match.addEventListener('change', onChange);
+
+		return () => {
+			match.removeEventListener('change', onChange);
+		};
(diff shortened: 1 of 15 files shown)


pandora-client-web/src/components/gameContext/notificationContextProvider.tsx
```diff
@@ -23,6 +23,9 @@ export const NOTIFICATION_KEY: Readonly<Record<NotificationSource, NotificationH
 	[NotificationSource.INCOMING_FRIEND_REQUEST]: null,
 };
 
+// TODO: Make it configurable whether or not to show popup
+const MOCK_SETTING_ENABLE_POPUP = false;
```
Sekkmer, line 27: I do not like the idea of turning these off, I often find it really useful to get this notification when I have Pandora in another tab, this is by no means a mock feature

Disabling also removes focus that we need to actually implement this really important feature

As for how it went through review, well, it is obviously something still under development, and only a dummy text is shown, though it could have said TODO

if you want to disable it please do it with a user-configurable server side setting

Sekkmer (changes requested):

Sekkmer (approved):

PR #304, opened 2023-06-18
[CHANGE] Room tab redesign

ClaudiaMia:
Redesigns the "Room" tab of room controls:
Preview

### DIFF
--- a/pandora-client-web/src/components/chatroom/chatroom.scss
+++ b/pandora-client-web/src/components/chatroom/chatroom.scss
@@ -1,5 +1,4 @@
-@import '../../styles/constants';
-@import '../../styles/utils';
+@import '../../styles/common';
 
 .chatroom {
 	overflow: hidden;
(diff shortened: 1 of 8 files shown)


pandora-client-web/src/components/chatroom/chatroom.scss
```diff
@@ -21,16 +20,55 @@
...
 	margin-bottom: spacing('large');
 
-	.safemode {
-		display: inline-flex;
-		background: $theme-warning;
-		border: solid black 1px;
-		padding: 0.25em;
-		align-items: center;
-		justify-content: center;
+	& fieldset {
```
Sekkmer, line 31: I think `&` is not required here, as in it changes nothing

pandora-client-web/src/components/wiki/pages/intro.tsx
```diff
@@ -41,10 +43,21 @@ export function WikiIntroduction(): ReactElement {
 				Your character's view can be toggled in the pose menu. The chosen position is seen by everyone in the room.
 			</p>
 
+			<h4>Persistent rooms</h4>
+			<p>
+				In Pandora, a room is always existing and has one or more owners. It is deleted only when it no longer has owners.
+				Each <strong>account</strong> has a limit on how many rooms it can own. You can see your account's owned rooms and limit in the room search screen.
```
Sekkmer, line 49: how about making it a bit more clear:

In Pandora, each room is persistent and is maintained by one or more owners. A room is only removed when it no longer has any owners. Every account has a maximum number of rooms it can own. You can view the rooms owned by your account, as well as your ownership limit, on the room search screen.

A room will appear in your room search only if it is public and occupied, or if you hold ownership or administrative rights over it. If you wish to create another room beyond your room limit, you must choose one of your existing rooms to either relinquish ownership (which may result in the room being deleted if there are no other owners), or repurpose it for your new needs.

Sekkmer (changes requested):

pandora-client-web/src/components/wiki/pages/intro.tsx
Jomshir98: I feel your proposal is better in some ways and worse in another. It does make things more clear, but it is also a lot harder to read, using precise, but "advanced" words a lot, also being a lot longer, which aren't good things for user-facing documentation, even less for the "intro" section that should only give brief overview of each important aspect.

Sekkmer: yeah probably somewhere in between would be better~~

Sekkmer: ok slightly shorter and less advanced words:

In Pandora, each room always exists and is managed by one or more owners. A room is removed only when it has no owners left. Your account has a certain limit for the number of rooms you can own. You can check out your rooms and your limit in the room search screen.

A room will show up in your search if it's either public and not empty or if you're an owner or admin of it. If you want to create a new room and you've reached your limit, you'll need to give up one of your current rooms. This could mean either giving up ownership (which might delete the room if there are no other owners) or repurposing it for your new needs.

ClaudiaMia: Thanks for your suggestion. I will cherry-pick parts of it into a complete rework of it I am doing together with Jomshir now~

Sekkmer (approved):

PR #311, opened 2023-06-19
[FEATURE] Show room background inside chatroom

Jomshir98:
The main purpose of this PR is to show room background inside chatroom, in same scale and offset to the target character as is visible in the room. Note, that it currently ignores room device position overrides.
Commits should be atomic and self-contained.

### DIFF
--- a/pandora-client-web/src/graphics/graphicsScene.tsx
+++ b/pandora-client-web/src/graphics/graphicsScene.tsx
@@ -146,6 +146,7 @@ export function GraphicsBackground({
 		backgroundAlpha: number;
 		backgroundImage: string;
 	}>(() => {
+		// Number is color in pixi format
 		if (typeof background === 'number') {
 			return {
 				backgroundTint: background,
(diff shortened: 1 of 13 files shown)


pandora-client-web/src/graphics/graphicsScene.tsx
```diff
@@ -177,21 +121,76 @@ function GraphicsSceneCore({
...
+				backgroundImage: '*',
+			};
+		}
+		if (/^#[0-9a-f]{6}([0-9a-f]{2})?$/i.test(background)) {
+			return {
+				backgroundTint: parseInt(background.substring(1, 7), 16),
+				backgroundAlpha: background.length > 7 ? (parseInt(background.substring(7, 9), 16) / 255) : 1,
+				backgroundImage: '*',
+			};
+		} else {
```
Sekkmer, line 169: nit: else could be removed as we return before

Sekkmer (approved):
Looks good

PR #312, opened 2023-06-19
[CHANGE] Reverse order of items in wardrobe

Jomshir98:
Not much to say, it makes things worn on top appear on top.


Sekkmer (approved):
I thought this would be configurable as a user setting, though it is much more simple to not allow it

But looks good and makes more sense reversed

Jomshir98:
I originally wanted to make it a setting, but I felt that introducing setting for something like this would make it far too confusing, especially when writing things in wiki or when people explain things to others. Having possibility of it being exact opposites depending on setting, would be too confusing to justify, especially as most non-coders agree on the new direction being more natural.

Issue #269, opened 2023-06-02
Split arm priorities


ClaudiaMia:
Currently, arms are on a single layer "ABOVE_ARMS" and "BELOW_ARMS". This leads to issues with more complex assets, where layers are either overlapping with itself in certain bone slider values or this is a limitation for the asset creator to show things how they are supposed to look in certain states.

This change should split the arms layer into 3 layers: hands, lower_arms, and upper_arms, which again should have variants for above and below, so a total of 6 priorities instead of 2 before.

Note: Items on the arms/hands also weirdly overlap with each other, e.g. mittens in pose "crossed front". Potentially this issue would also address this.

Jomshir98:
This change would strongly benefit from [pandora-assets#37](https://github.com/Project-Pandora-Game/pandora-assets/issues/37), as adding more layers would make it bothersome to do this without some automation to help.

ClaudiaMia:
Two more aspects that should be investigated:

- if there is a need for more hand turns/views, see Lily's comment in #general on 6-11-2023 or if this would make asset creation covering hands too much hell.
- potentially if we can angle the hand to some small degree around the wrist joint and make it look good enough

PR #313, opened 2023-06-20
[CHANGE] Graphics: Split left and right arm priority

Jomshir98:
This change allows explicitly differentiating between left and right arm. This prevents strange clipping when left and right hand overlap, for example hands seemingly going into other arm's sleeve with long shirt.

This is related to #269, but doesn't solve it fully, only one of the issues there, namely:
> Items on the arms/hands also weirdly overlap with each other, e.g. mittens in pose "crossed front".


pandora-client-web/src/graphics/def.ts
```diff
@@ -24,19 +31,27 @@ export const COMPUTED_LAYER_ORDERING = [
...
-	'BELOW_ARMS_BACK',
-	'ARMS_BACK',
-	'ABOVE_ARMS_BACK',
+	'BELOW_ARM_LEFT_BACK',
+	'ARM_LEFT_BACK',
+	'ABOVE_ARM_LEFT_BACK',
+
+	'BELOW_ARM_RIGHT_BACK',
+	'ARM_RIGHT_BACK',
+	'ABOVE_ARM_RIGHT_BACK',
```
Sekkmer, line 40: not sure how simple it would be but it might look more readable if this is generated by a function, same for the other sections that got duplicated

pandora-client-web/src/graphics/def.ts
```diff
@@ -56,14 +71,20 @@ export const PRIORITY_ORDER_SPRITES = [
...
+	'ABOVE_ARM_LEFT_BACK',
+	'ABOVE_ARM_LEFT_FRONT',
+	'BELOW_ARM_RIGHT_BACK',
+	'BELOW_ARM_RIGHT_FRONT',
+	'ARM_RIGHT_BACK',
+	'ARM_RIGHT_FRONT',
+	'ABOVE_ARM_RIGHT_BACK',
+	'ABOVE_ARM_RIGHT_FRONT',
 	'FRONT_HAIR',
 	'ABOVE_FRONT_HAIR',
```
Sekkmer, line 87: is this intentional, to put these above the arms?

(wouldn't ask if it were a separate commit)

pandora-common/src/assets/graphics/graphics.ts
```diff
@@ -143,9 +143,12 @@ export const LAYER_PRIORITIES = [
 	'ABOVE_BODY',
 	'FRONT_HAIR',
 	'ABOVE_FRONT_HAIR',
```
Sekkmer, line 145: like here the order is not swapped

pandora-common/src/validation.ts
```diff
@@ -46,6 +46,15 @@ export function ArrayToTruthyMap<T extends string>(array: readonly T[]): Record<
 	return result;
 }
 
+/**
+ * Checks whether given array includes the element, working as a guard
+ * @param array - The array to check in
+ * @param element - Element to find
+ */
+export function ArrayIncludesGuard<T extends string | number | boolean | null | undefined>(array: readonly T[], element: unknown): element is T {
```
Sekkmer, line 54: you could probably add `symbol` though not required

Sekkmer (changes requested):

pandora-client-web/src/graphics/def.ts
Jomshir98: I prefer intentionally leaving this more explicit, as it still is short and easy to validate by TS this way. It also to me does feel easier to read honestly, than introducing functions.

pandora-client-web/src/graphics/def.ts
Jomshir98: Yes, done to bring this layering more closer to the normal layering. By this point I'm thinking of just dropping the sprite order, as it doesn't really help anything anymore. Its original purpose was simplified layering when sprites didn't properly interleave in old demos, which I think changed with react-pixi rewrite.

pandora-common/src/assets/graphics/graphics.ts
Sekkmer: is this good too, to not swap the order here

Sekkmer (approved):

PR #316, opened 2023-06-23
[FEATURE] password lock

Sekkmer:


ClaudiaMia (changes requested):
I was testing this with what was deployed on the official server and have the following comments:

- The input field and the lock button overlap. There should be a normal amount of space between them.
- Not sure this is an issue of this PR, but I saw some related changes in the code, so maybe it is an issue of this PR: When you are in safemode and in the wardrobe and try to leave safemode with the context menu on top, you are forwarded to a blank page with a fatal
`[ErrorBoundary] Uncaught error
 NotFoundError: Node.insertBefore: Child to insert before is not a child of this node @ https://project-pandora.com/index.3ac7f1adbeac3f1e9574.js:2:0
 DOMException: Node.insertBefore: Child to insert before is not a child of this node`
- issue with option "using old password": lock any lock with an input; unlock it (without the option to remove the password); press option to reuse the old password; lock it; it does not show as locked but IS actually locked; you are stuck as you cannot enter any password to unlock it

Sekkmer:
Fixed all 3 in a separate atomic commit so it should be simple to review the changes

The error was created by #315 and it was also effecting the color dialog

Sekkmer:
unlock issue is fixed

ClaudiaMia (approved):
Can confirm that all issues are fixed now. Great work!

PR #325, opened 2023-07-03
[FEATURE] device pose preset and exit pose

### LINKED ISSUE #283: Device-specific pose template extension
Sekkmer: new idea, what about device pose
Sekkmer: like you could get extra set of poses when you are inside the device
Sekkmer: kind of like your current pose
Claudia: Ohh~
Claudia: so basically, it simply adds some new poses (maybe in a different color, button-wise) to the other poses listed under the pose-tab
Claudia: that are device specific
Sekkmer: mhm
Claudia: Yeah, I like that idea, as this current pose that I added to the universal template
Claudia: makes not much sense outside of specific chairs with an arm rest

Sekkmer:
resolve #283

### DIFF
--- a/pandora-client-web/src/assets/appearanceValidation.ts
+++ b/pandora-client-web/src/assets/appearanceValidation.ts
@@ -79,6 +79,8 @@ export function RenderAppearanceActionResult(assetManager: AssetManagerClient, r
 				return `The ${DescribeAsset(assetManager, e.asset)} cannot be added, removed, or modified, because ${DescribeAssetSlot(assetManager, e.slot)} is covered by another item.`;
 			case 'blockedHands':
 				return `You need to be able to use hands to do this.`;
+			case 'exitPose':
+				return `You cannot assume the pose required to exit the ${DescribeAsset(assetManager, e.asset)}.`;
 			case 'invalid':
 				return '';
 		}
--- a/pandora-client-web/src/common/useEvent.ts
+++ b/pandora-client-web/src/common/useEvent.ts
@@ -1,29 +1,25 @@
 import { useRef, useLayoutEffect, useCallback, useState } from 'react';
 import { useMounted } from './useMounted';
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-type AnyFunction = (...args: any[]) => any;
-
 /**
  * Creates a stable function that wont change during the lifecycle of the component.
  * @param callback - The function to memoize.
  * @see https://github.com/reactjs/rfcs/blob/useevent/text/0000-useevent.md
  */
-export function useEvent<T extends AnyFunction>(callback: T): T {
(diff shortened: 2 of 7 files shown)


ClaudiaMia (commented):
I checked it out on Jomshir's server and also linked the asset repo to it, but I am unable to use the new functions exitPose and posePresets, as they are not known.

Sekkmer:
Sure, I make an assets repo PR as well

Sekkmer:
@ClaudiaMia I deployed this branch with the heart throne changes

pandora-client-web/src/assets/appearanceValidation.ts
```diff
@@ -79,6 +79,8 @@ export function RenderAppearanceActionResult(assetManager: AssetManagerClient, r
 				return `The ${DescribeAsset(assetManager, e.asset)} cannot be added, removed, or modified, because ${DescribeAssetSlot(assetManager, e.slot)} is covered by another item.`;
 			case 'blockedHands':
 				return `You need to be able to use hands to do this.`;
+			case 'exitPose':
```
ClaudiaMia, line 82: Is this considered to be a debugging fail-safe in case some asset creator specified an exit pose that is impossible to assume under certain conditions?

What happens in this case? They cannot exit the device? Can someone else exit the character or are they stuck?

ClaudiaMia (approved):

pandora-client-web/src/assets/appearanceValidation.ts
ClaudiaMia: ... or maybe someone adds new items to the character while inside the slot that makes one unable to exit the device? Like making on unable to stand up from a chair.

Sekkmer: There are currently no checks, so you can create an asset that is impossible to exit.

We could introduce an extra check later that ensures you can assume the exit pose before entering a device. This way, if the exit pose is impossible, you can not even enter.
 
 `optional:` part of the preset can always be used as a convenience, like assume a regular arm pose exiting a cross, without being a requirement.
 
 And the intended usage for the required exit pose is indeed to allow assets restricting the pose, like sitting to keep the player tied down to a chair

Sekkmer: Added the missing safemode check, so even if the asset is faulty, you can exit it in safemode

ClaudiaMia: Great, thanks for explaining.
Yes, this makes sense to me.
I think it is fine to not check if you can exit when enter, since things can change while inside the slot, and with your last change, wrongly defined assets can be left with safemode in the worst case, if we cannot spot such things during reviews or tests.

ClaudiaMia (approved):

Issue #70, opened 2022-11-14
Persistent rooms


Jomshir98:
See:
https://discord.com/channels/872284471611760720/1025099479877169233/1025099544205209690

@SandrinePDR :
I'll try to add a list of tasks that need to be done to solve this (from my point of view). Feel free to add/remove further entries or group the entries in a sensible ways. If it is possible to create sub-issues, we should convert the list to those. I didn't do this now, because I did not want to spam our general issue list.

As the list got quite long, perhaps @Jomshir98 should have a look at it and sort the topics by being relevant to the MVP

### Relevant for the MVP
- [x] Define the room creation workflow 
- [x] Add an upper limit for the rooms (areas?), an account can own
- [x] Add something like a room capacity, so that a room cannot be spammed by too many items 
([Claudia]: I believe this point will be implemented by https://github.com/Project-Pandora-Game/pandora/issues/3)
- [x] Add a 'capacity-taking' attribute to assets, where most assets should have a value of 0 ([Claudia]: I believe this point will be implemented by https://github.com/Project-Pandora-Game/pandora/issues/3 and likely https://github.com/Project-Pandora-Game/pandora/issues/19)

### Can be done later

- [x] Add a private room for every new character on character creation
- [x] Move characters to their private room upon leaving/ dc-ing
- [ ] Add 'Lobby' room to a shard
- ~~Make 'special chars' e.g. lovers/owner(s?) admins of that private room~~
- [ ] Show the name of the room, a character is in in the overview screen
- ~~Add owner- and VIP-lists to the room attributes~~ (?)
- [x] Delete private rooms on character deletion
- [x] Make an empty/ afk room go to sleep
- [ ] Add an option to a character to make them enter their personal room on disconnect (time customizable)
- [x] Add a beep/ invitation option
- [ ] Make other chars enter a room on accepting an invitation
- [x] Have a VIP-list (add/ remove other accounts (chars?))
- [x] Have a black-list (add/ remove other accounts (chars?))
- [x] Destroy a room upon last owner leaving 
([Claudia]: I believe it is only an option. The default would be that just gets delisted if it was public. Moreover we need: Destroy a room upon an owner deleting it or reusing the slot)
- [x] Add a 'room construction toolkit' item
- [ ] Hide inactive chars that are able to move freely (e.g. not part of the room, by being in a cage, chained to the floor, etc.) [Claudia: Is this still needed? I feel that could be confusing/untransparent for users]
[...]

PR #334, opened 2023-08-02
[FEATURE] Chatroom offline character persistence

Jomshir98:
Works towards: #70 
Based on following proposal: [Pandora Discord channel link](https://discord.com/channels/872284471611760720/1025099479877169233/1121746790849449985)

This PR implements part of the proposal, to be specific it does the following:
- Characters now always stay in rooms unless they leave manually
- Characters and rooms get properly loaded and unloaded as needed, if room is loaded, it always loads all characters that are inside
- Fixed bug with directory could crash when stopping with connected shards
- Updated displaying online characters (both in contacts and in own character list) to only show when the character is actually selected by player
- While at it fixed bug that new friends show as offline in the list until refresh
- Show "Offline" status next to offline characters in the "Room" tab character list
- Show offline characters differently in graphics (there is a new "Interface" option to change what effect is used, defaults to "Ghost" (darken + semi-transparent)
- Added a warning to the leave dialog, if inside a room, that character will stay in it even on character change/logout
- Fixed characters being removed from room devices when room unloads or is moved to another shard

Known bugs:
- ~~Characters are still randomly removed from room devices when the room gets unloaded~~ _fixed_

Related things NOT implemented as part of this, will be done in later PRs (unless we decide to delay this until said feature is done):
- Timer to auto-kick/auto-leave the room when offline - characters are only removed based on manual action
- Chat messages storing logic for offline characters (currently offline characters do not store messages at all)
- Any extra chat messages around disconnecting/reconnected characters


pandora-server-directory/src/database/mongoDb.ts
```diff
@@ -325,7 +359,7 @@ export default class MongoDatabase implements PandoraDatabase {
 		return await this._chatrooms.find({
 			owners: { $elemMatch: { $in: [account] } },
 		})
-			.project<Pick<IChatRoomDirectoryData, 'id' | 'config' | 'owners'>>({ id: 1, config: 1, owners: 1 })
+			.project<Pick<IChatRoomDirectoryData, 'id' | 'config' | 'owners' | 'accessId'>>({ id: 1, config: 1, owners: 1, accessId: 1 })
```
Sekkmer, line 362: can we use `typeof CHATROOM_DIRECTORY_PROPERTIES[number]` here?

pandora-client-web/src/components/chatroom/chatRoomCharacter.tsx
```diff
@@ -124,23 +127,35 @@ function ChatRoomCharacterDisplay({
...
+		position: dataPosition,
+		settings,
+		isOnline,
+	} = useCharacterData(character);
+
+	const { interfaceChatroomOfflineCharacterFilter } = useCurrentAccountSettings();
+
+	const playerFilters = usePlayerVisionFilters(character.isPlayer());
+	const characterFilters = useCharacterDisplayFilters(character);
+	const filters = [...playerFilters, ...characterFilters];
```
Sekkmer, line 142: shouldn't this be a `useMemo`?

Sekkmer (changes requested):

pandora-server-directory/src/account/accountRelationship.ts
```diff
@@ -85,9 +87,11 @@ export class AccountRelationship {
 		}
 		if (this.account.data.settings.allowDirectMessagesFrom === 'room') {
 			for (const char of this.account.characters.values()) {
-				if (!char.room) continue;
+				const room = char?.loadedCharacter?.room;
```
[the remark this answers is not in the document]
Jomshir98: I'm not quite sure what you mean? To me it is better keeping only one source of truth, even if it means checking slightly more data than needed and in this case it feels minimal - from the old behaviour it changed only by adding the middle step of `loadedCharacter`. Also I'm not sure if the `char` can actually ever be null or if the nullcheck is there only to stop TS from complaining.

pandora-server-directory/src/account/character.ts
```diff
@@ -155,208 +99,396 @@ export class Character {
...
+		return Promise.resolve();
+	}
+}
+
+export type CharacterAssignment = {
+	// Character is assigned to a specific shard and is not in any room
+	type: 'shard';
+	shard: Shard;
+} | {
+	// Character is assigned to the same shard the room is, but is not in the room
```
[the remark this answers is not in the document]
Jomshir98: This is described a bit obscurely in room join:
> Must connect to the same shard as the room to check character-based join requirements
I will add comment that this is necessary for the flow of joining a room.

pandora-server-directory/src/database/mongoDb.ts
Jomshir98: I would agree with doing that if the pick filter object would also be autogenerated. I stuck with doing it manually as then TS catches that (for mongo the object can be any record, it doesn't check the exact property names, making it unsafe without being this explicit)

pandora-client-web/src/components/chatroom/chatRoomDevice.tsx
```diff
@@ -375,11 +375,13 @@ function RoomDeviceGraphicsLayerSlot({ item, layer }: {
...
 	layer: IRoomDeviceGraphicsLayerSlot;
-	character: Character;
+	character: Character<ICharacterRoomData>;
 	characterState: AssetFrameworkCharacterState;
 }): ReactElement | null {
 	const debugConfig = useDebugConfig();
-	const filters = usePlayerVisionFilters(character.isPlayer());
+	const playerFilters = usePlayerVisionFilters(character.isPlayer());
+	const characterFilters = useCharacterDisplayFilters(character);
+	const filters = [...playerFilters, ...characterFilters];
```
Sekkmer, line 384: here too, I just forgot to mark it last time

Sekkmer (approved):

PR #368, opened 2023-08-20
[FEATURE] links in chat messages

Sekkmer:
parse `http://` and `https://` as raw occ messages, automatically parse as raw when inside ooc
parse links inside raw ooc messages and display them in an anchor element


ClaudiaMia (approved):
LGTM

PR #372, opened 2023-08-23
[CHANGE] Cleanup contacts list, add online indicator

Jomshir98:
This PR does following changes:
- Splits `relationships.tsx` file into this and context file, so UI can be changed without whole page reloading (no changes to the context code, except rename of one function)
- Cleanups style of contacts list by sorting columns differently and making some smaller
- Renames column "Created" to "Since" and only shows date instead of date and time
- Adds a green or empty dot next to "Online" or "Offline", respectively, for quicker orientation; also makes the Offline text into italics


ClaudiaMia (approved):
LGTM!

Issue #10, opened 2022-08-29
E2E tests


Jomshir98:
There is missing architecture for tests that test interactions between individual projects - like automatic tests of integration and communication between Client, Directory and Shard, where none of the components are mocked.

PR #375, opened 2023-08-27
[DEV] Basic framework for End-to-End testing

Jomshir98:
Resolves: #10 

This PR adds a framework for writing E2E tests. Currently it doesn't do much, but it does a bit, namely:
- Uses Playwright for using a full browser for testing (currently configured to use stable chromium, more browsers are supported; I chose Playwright over Puppeteer in the end as it has much better support for actually writing tests)
- Builds common, servers and client in a way usable for tests
- Runs a simple http server (using `superstatic`) that supports path rewrites we need for client
- Can start Directory server with custom env (no Shard support yet)
- Detects if client crashes or produces error (no server crash/log detection during test yet)
- Has several basic tests
  - Client loads both main client and Editor landing page without error
  - Some EULA tests
  - Test about Client connecting and reconnecting to Directory
- Adds a `test-e2e` global script for running tests locally, also adds this to the `ci` global script
- Adds setup for GitHub actions

Additional notes:
- The setup is done in such a way that slowdown unless actually using E2E tests is minimal. Browsers for testing are only downloaded when needed
- E2E testing CI produces a HTML report artifact together with a basic text-based output of passing tests
- There are two very small changes to client code to make this work; no changes to server (but some will be needed to get registration testing possible with emails)
- E2E tests are under GPL license (same as servers and client)


pandora-tests/test/utils/server.ts
```diff
@@ -0,0 +1,141 @@
+import { Assert, AssertNotNullable } from './utils';
+import { ChildProcessWithoutNullStreams, spawn } from 'child_process';
+import { TEST_DIRECTORY_PORT, TEST_SERVER_DIRECTORY_ENTRYPOINT, TEST_SERVER_DIRECTORY_TEST_DIR, TEST_TEMP } from '../_setup/config';
+import { test } from '@playwright/test';
+import path from 'path';
+
+/* eslint-disable @typescript-eslint/naming-convention */
+interface DirectoryEnvSetup {
```
Sekkmer, line 8: is there any way to not have a duplicate of this, like what if someone adds a new config to directory but forgets to add it here?

pandora-tests/test/utils/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+		throw new Error(msg ? `Assetion failed: ${msg}` : 'Assertion failed');
+	}
+}
+
+/**
+ * Assert all arguments are `never`
+ *
+ * Useful for checking all possible outcomes are handled
+ */
+export function AssertNever(...args: never[]): never {
```
Sekkmer, line 12: wouldn't it be better to use these from common, or what is the reason not to import that?

Jomshir98:
At the moment answer to both is, that building of common/servers happens as part of actual E2E pipeline - its type-checking happens before common and servers are built. This is surely something that could be improved later; right now this was the simplest solution to this.

pandora-tests/test/utils/server.ts
Jomshir98: Created an issue #376 for this

PR #400, opened 2023-09-18
[DEV] Fix race condition in automation task

Jomshir98:
Example of this can be seen in #399.


Jomshir98:
Or here... (caused by quick addition of labels)

Sekkmer (approved):

PR #401, opened 2023-09-19
[FIX] server crash on client reconnect

### LINKED ISSUE #378: Server crash: Race condition while assigning character on client reconnect
Commit the stack is from: a922f29e57f8f30447a36beb92a09be0a7e84bf2
During client reconnection the following error can happen.
```
Error processing connect auth from *** [Error: Assertion failed
    at Assert (pandora/pandora-common/src/utility.ts:68:9)
    at ClientConnection.setCharacter (pandora/pandora-server-directory/src/networking/connection_client.ts:86:10)
    at ConnectionManagerClient.handleAuth (pandora/pandora-server-directory/src/networking/manager_client.ts:496:17)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
]
```
Cause of this is trying to reconnect to a character that already has a client assigned. To fix this an assigned client should be removed first if the connect secret matches and then the new client should be assigned to the character, without changing the connect secret.
This will result in similar flow to `pandora-server-directory/src/account/character.ts > Character.connect()`, but without generating new `_connectSecret` or calling `this.currentShard.update()`.
This can be simply implemented as an argument to `Character.connect()` not to update secret if one already exists and replacing the usage in `handleAuth`.

ClaudiaMia:
No idea how to test if what I did is correct or fixes it. But I can still connect and reconnect and switch characters!

closes #378

### DIFF
--- a/pandora-server-directory/src/account/character.ts
+++ b/pandora-server-directory/src/account/character.ts
@@ -277,7 +277,7 @@ export class Character {
 	//#region Client connection handling
 
 	@AsyncSynchronized('object')
-	public async connect(connection: ClientConnection): Promise<'ok' | 'noShardFound' | 'failed'> {
+	public async connect(connection: ClientConnection, newConnection: boolean = true): Promise<'ok' | 'noShardFound' | 'failed'> {
 		if (this._disposed)
 			return 'failed';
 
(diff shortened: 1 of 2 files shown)


Jomshir98 (approved):
LGTM!

PR #416, opened 2023-09-27
[FIX] Hotfix: An important line got lost somewhere

Jomshir98:
Unfortunately `pull_request_target` CI cannot be tested differently than having it somewhere on `master`...


ClaudiaMia (approved):
LGTM!

Issue #376, opened 2023-08-27
Server configuration framework


Jomshir98:
At the moment our server configuration happens simply by reading values from environment into constants in a `config.ts` file.
This is starting to become problematic as more and more are added and we want more features. For that reason it would be beneficial to find or make a framework for defining configurations.

Properties of such system that would be beneficial for us:
- Allows validating data during load, so any use site doesn't have to do its own validation
- Makes it possible to change the values without restarting the server (e.g. configuration to block registration/password resets, list of admin accounts, ...)
- Configs should be editable offline too - even before server is started for the first time
- Configs should be passable through environment as it is currently; mainly important for shards which should only require configuring through docker-compose service and then simply work
- Makes it possible to create a strict typings of configuration that can be used by for example tests

PR #418, opened 2023-10-04
[FEATURE] zod env

Sekkmer:
works towards #376


pandora-tests/test/utils/server.ts
```diff
@@ -86,14 +37,35 @@ const DIRECTORY_ENV_DEFAULTS: DirectoryEnvSetup = {
 export interface TestStartDirectoryOptions {
 	/** Whether to keep the server active across tests */
 	keepActive: boolean;
-	configOverrides: Partial<DirectoryEnvSetup>;
+	configOverrides: DirectoryEnvSetup;
```
Jomshir98, line 40: This should still be only partial, as it overrides the already complete set in `DIRECTORY_ENV_DEFAULTS`.

pandora-tests/test/utils/server.ts
```diff
@@ -4,68 +4,19 @@ import { TEST_DIRECTORY_PORT, TEST_SERVER_DIRECTORY_ENTRYPOINT, TEST_SERVER_DIRE
...
-	/** Path to HTTPS certificate file */
-	SERVER_HTTPS_CERT: string;
-	/** Path to HTTPS key file */
-	SERVER_HTTPS_KEY: string;
-	/** Secret key used to authenticate Shards connecting to Directory */
-	SHARD_SHARED_SECRET: string;
-	/** How many hops are we after a trusted reverse proxy */
-	TRUSTED_REVERSE_PROXY_HOPS: string;
+import type { ENV } from 'pandora-server-directory/src/config';
+import { EnvStringify } from 'pandora-common/dist/environment';
```
Jomshir98, line 8: Nit: The `EnvStringify` might be better re-exported through `utils` file.

pandora-server-shard/src/config.ts
```diff
@@ -1,61 +1,63 @@
...
+
+	/** The name of the application */
+	APP_NAME: z.string().default('Pandora Shard Server'),
+
+	/** If this shard is in development mode */
+	SHARD_DEVELOPMENT_MODE: z.boolean().transform((arg) => arg || process.argv.includes('--development')).default(false),
+
+	//#region Networking
+	/** Port on which will HTTP server listen */
+	SERVER_PORT: z.number().default(25561),
```
Jomshir98, line 22: Nit: This should also specify it should be int, to preserve the removed check
```suggestion
	SERVER_PORT: z.number().int().default(25561),
```

pandora-server-directory/src/config.ts
```diff
@@ -1,102 +1,112 @@
...
-/** How many hops are we after a trusted reverse proxy */
-export const TRUSTED_REVERSE_PROXY_HOPS = Number.parseInt(process.env.TRUSTED_REVERSE_PROXY_HOPS ?? '0') || 0;
+	APP_NAME: z.string().default('Pandora Directory Server'),
 
-//#endregion
+	//#region Networking
 
-//#region Logging
+	/** Port on which will HTTP server listen */
+	SERVER_PORT: z.number().default(25560),
```
Jomshir98, line 11: Nit: This should also specify it should be int, to preserve the removed check
```suggestion
	SERVER_PORT: z.number().int().default(25560),
```

Jomshir98 (approved):
Looks good! Just a few nitpicks~

PR #423, opened 2023-10-09
[FEATURE] Assets loader automatic retry

Jomshir98:
This PR implements automatic retry of loading resources that failed to load. It also moves in direction of being able to unload resources when we no longer need them, but that part is not yet implemented.
The actual loading is complicated, because PIXI has its own Texture cache, that however keeps textures that failed to load, too.


pandora-client-web/src/graphics/useTexture.ts
```diff
@@ -3,50 +3,49 @@ import { useEffect, useRef, useState } from 'react';
...
-		wantedGetTexture.current = getTexture;
-		if (!getTexture || image === '') {
-			setTexture({
-				image,
-				texture: Texture.EMPTY,
-			});
+		const loader = manager?.loader;
+		if (customGetTexture || loader == null || SPECIAL_TEXTURES.has(image)) {
+			wanted.current = '';
+			setTexture(RESULT_NO_TEXTURE);
```
Sekkmer, line 44: how does this return `'*', Texture.WHITE`?

Jomshir98: I split that logic into more generic `SPECIAL_TEXTURES` map, which is then used outside of this async getter to avoid waiting a frame to load these special textures (as the effect is async, it wouldn't resolve until next frame).
The code that handles these overrides is on lines 79-81.

pandora-client-web/src/assets/graphicsLoader.ts
```diff
@@ -1,8 +1,11 @@
 import { Assert, GetLogger, Logger } from 'pandora-common';
-import { BaseTexture, IImageResourceOptions, Resource, Texture, autoDetectResource } from 'pixi.js';
+import { BaseTexture, IImageResourceOptions, Resource, Texture, autoDetectResource, settings as PIXISettings } from 'pixi.js';
 import { PersistentToast } from '../persistentToast';
 import { IGraphicsLoader } from './graphicsManager';
 
+// Prevent PIXI from doing PIXI things
+PIXISettings.STRICT_TEXTURE_CACHE = true;
```
Sekkmer, line 7: hmm could we put this in a file on its own, I remember we had that batch-size thing we no longer need and that was kind of hard to find, so maybe a ts file with a good name could help

Sekkmer (approved):

PR #425, opened 2023-10-15
[FIX] Fix alphamask change sometimes not trigerring redraw

Jomshir98:
Also avoids crashing when the mask is set to empty (which was why the original code was written in the way it was).
Adds a small fix to fix React's warning about duplicate key in wardrobe expressions, when two items have modules with same name.


ClaudiaMia (approved):
Works!

Issue #110, opened 2022-12-06
Better startup/cleanup


Sekkmer:
currently, we have many classes with `.init()` methods, and some also have `.onDestroy,` but thy lifecycle is not well defined

it would be nice to refactor this to something more predictable

we should call each `.onDestroy` in reverse of the init

something similar to this:
(where add could accept a callback with the service to log we are done with a step)

```ts
interface Service {
	inti(): Promise<Service> | Service;
	onDestory(): void;
}

class ServiceBuilder {
	private readonly services: Service[] = [];

	public add(service: Service): this {
		this.services.push(service);
		return this;
	}

	public async build(): Promise<void> {
		for (const service of this.services) {
			await service.inti();
		}
	}

	public destory(): void {
		for (const service of this.services) {
			service.onDestory();
		}
	}
}
```

PR #427, opened 2023-10-16
[REFACTOR] unify Service base type

Sekkmer:
working towards #110


pandora-common/src/services/serviceManager.ts
```diff
@@ -0,0 +1,4 @@
+export type Service = {
+	init?(): Promise<Service> | Service;
+	onDestroy?(): Promise<void> | void;
```
Jomshir98, line 3: Nit: Maybe `destroy` would be better? Having an "on" prefix signifies that you should never call this method yourself - it happens as response to some event... Hmm~ I guess I feel neutral, as it kinda is reaction to server being destroyed... but maybe then `onShutdown` might suit better?

pandora-server-directory/src/lifecycle.ts
```diff
@@ -17,25 +17,35 @@ const logger = GetLogger('Lifecycle');
 	wtfnode.setLogger('error', (...message) => wtfNodeLogger.error(...message));
 }
 
+let destroying: string | undefined;
 let stopping: Promise<void> | undefined;
 const STOP_TIMEOUT = 10_000;
 
+function DestroyService(service: Service & object): Promise<void> | void {
```
Jomshir98, line 24: I don't think `& object` needs to be here
```suggestion
function DestroyService(service: Service): Promise<void> | void {
```

Jomshir98 (changes requested):

pandora-common/src/services/serviceManager.ts
Sekkmer: well we did have onDestroy for most of the cases, so that's why I went with it

pandora-server-directory/src/database/databaseProvider.ts
```diff
@@ -2,12 +2,12 @@ import { MockDatabase } from './mockDb';
 import MongoDatabase from './mongoDb';
 import { ENV } from '../config';
 const { DATABASE_TYPE } = ENV;
-import type { CharacterId, IChatRoomData, ICharacterData, ICharacterDataAccess, ICharacterSelfInfo, ICharacterSelfInfoUpdate, IDirectoryAccountSettings, IDirectoryDirectMessage, IDirectoryDirectMessageInfo, IChatRoomDataDirectoryUpdate, IChatRoomDataShardUpdate, RoomId, IChatRoomDirectoryData, AccountId } from 'pandora-common';
+import { type CharacterId, type IChatRoomData, type ICharacterData, type ICharacterDataAccess, type ICharacterSelfInfo, type ICharacterSelfInfoUpdate, type IDirectoryAccountSettings, type IDirectoryDirectMessage, type IDirectoryDirectMessageInfo, type IChatRoomDataDirectoryUpdate, type IChatRoomDataShardUpdate, type RoomId, type IChatRoomDirectoryData, type AccountId, type Service } from 'pandora-common';
```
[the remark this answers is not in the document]
Sekkmer: not sure why VSCode decide to expand this... maybe I imported something then removed it

Jomshir98 (approved):
Looks good to me! I added only one small commit to avoid unnecessary cast.

PR #430, opened 2023-10-18
[FEATURE] Character logic framework and initial permissions implementation

Jomshir98:
This PR:
- Refactors some character data passing around asset framework into proper classes that have different implementations on client and server
- Refactors appearance action context passing
- Allow appearance actions to report multiple problems; not just the first one
- Adds base framework for interactions and permissions (permissions for now are simple toggle "allow others" yes/no)
- Adds configuration screen for permissions into settings
- Adds a toast when wardrobe action is rejected on the server for any reason
- Adds a check where wardrobe queries the server for having permission and the result is checked (and invalidated when someone in room changes their permissions)

Full permission config is considered private information and only Player herself can get it currently. Checking is done by querying server with format "Do I have permission xyz for character c?".
When designing the framework for interactions and permissions I focused on making it easy to add new interaction types and such new permissions - doing that only needs editing the interaction config and then using that interaction using its id wherever needed during appearance action.
Permissions have default values. If player never touches a permission (or resets it to the default) and the default value changes, then they do get new defaults. If player, however, has different value that is default, their setting stays even if default changes.


pandora-client-web/src/components/wardrobe/wardrobeContext.tsx
```diff
@@ -127,43 +111,75 @@ export function useWardrobeContext(): Readonly<WardrobeContext> {
...
-	return useAsyncEvent(async () => {
-		if (action)
-			return await execute(action);
-
-		return null;
-	}, ExecuteCallback(props));
+	return useAsyncEvent(
+		async (action: AppearanceAction) => await execute(action),
+		(result) => {
+			{
```
Sekkmer, line 123: you can remove this extra block

Sekkmer (approved):

Issue #432, opened 2023-10-21
Limit all user inputs to reasonable values


ClaudiaMia:
The goal is to prevent the scenario that someone can flood the database by being able to write huge amounts of data to it.

To that end, all input fields of the client shall be limtied to reasonable character limits and validated on the server-side.

That incldues:

- [ ] Account and character creation
- [ ] room name and description
- [ ] item template name and description (not yet implemented)
- [ ] direct messages
- [ ] chat messages
- [ ] account and character profile (not yet implemented)

Sekkmer:
Since DMs are encrypted using AES-GCM 256, and then Base64 encoded, whatever character limit we set should be increased by roughly `35%` at least when checked by directory.

PR #435, opened 2023-10-28
[CHANGE] Login preview, input limiting, and chatroom preview

ClaudiaMia:
Adds input validation constants as listed by #432 where already possible.
Does not only limit the input on client and server side but also shows information about current and maximum values to the user.

Additionally some more smaller improvements:
- Chatroom list supports different screen sizes better
- Chatroom preview has more space to show the description by default
- Login preview has one more page with new text and supports external images now
- Wiki includes permission section


pandora-client-web/src/config/definition.ts
```diff
@@ -3,6 +3,7 @@ import { z } from 'zod';
 export const WEBPACK_BASE_CONFIG = {
 	DIRECTORY_ADDRESS: z.string().default('http://127.0.0.1:25560'),
 	EDITOR_ASSETS_ADDRESS: z.string().default('http://127.0.0.1:26969/assets'),
+	EXTRA_ASSETS_ADDRESS: z.string().default(''),
```
Sekkmer, line 6: What exactly is the plan for this, also where the preview images, in assets repo?

ClaudiaMia: /me explained the idea to Sekkmer:
Sekkmer: it could be more controlled though
Sekkmer: like what if an image is missing
Sekkmer: what if it fails to load

Jomshir98: @Sekkmer:
It was created with following points in mind:
- We don't want fully rendered scenes in any public repo, to adhere to TOS
- The images need to be served before login - so before selecting character too, meaning they can't be part of asset handling pipeline
- There shouldn't be any error produced when running local development client

This way of handling it felt as the most straightforward one:
It allows specifying URL where to load things from (can be the same as public assets url, but doesn't have to be - it is dependent entirely on how we do deployment; right now I just updated the NGinx setup to host it on a specific path).
I believe we don't need some handling for when this fails to load - it would behave the same as any other image on the internet that fails to load. It would also retry by itself the next time the image is focused while the login preview is cycling around.

Sekkmer (approved):

PR #439, opened 2023-11-02
[FIX] minor bugs

Sekkmer:
Fixes:
 - null dereference if invalid DM is selected
 - initial fixup on permission is not saved

remove `axios` as `fetch` is now stable in node

minor UI improvement on not found account for DMs

### DIFF
--- a/pandora-client-web/src/components/directMessage/directMessage.tsx
+++ b/pandora-client-web/src/components/directMessage/directMessage.tsx
@@ -81,6 +81,7 @@ function DirectMessageElement({ message, channel, account }: { message: DirectMe
 
 function DirectChannelInput(): ReactElement | null {
 	const channel = useDirectMessageChannel();
+	const ref = React.useRef<HTMLTextAreaElement>(null);
 
 	const onKeyDown = useEvent((ev: React.KeyboardEvent<HTMLTextAreaElement>) => {
 		const textarea = ev.currentTarget;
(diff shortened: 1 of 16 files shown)


pandora-client-web/src/components/directMessage/directMessage.tsx
```diff
@@ -94,8 +96,17 @@ function DirectChannelInput(): ReactElement {
 		}
 	});
 
+	React.useEffect(() => {
+		ref.current?.focus();
+	}, [channel.account]);
```
Jomshir98, line 101: [non-blocking] We might want to create a hook similar to what we have in `chatInput.tsx` "Handler to autofocus chat input": Hook that listens to global keypresses and focuses the input if no other input is focused.

pandora-client-web/src/components/directMessages/directMessages.tsx
```diff
@@ -76,17 +79,25 @@ function DirectMessageInfo({ info, selected }: { info: Readonly<IDirectoryDirect
...
+		const parsed = parseInt(ref.current.value, 10);
 		if (Number.isInteger(parsed) && parsed > 0 && parsed !== accountId) {
 			directory.directMessageHandler.setSelected(parsed);
 		}
-	});
+	}, [directory.directMessageHandler, accountId]);
+	const onKeyDown = React.useCallback((e: React.KeyboardEvent<HTMLInputElement>) => {
+		if (e.key === 'Enter') {
+			onClick();
+		}
```
Jomshir98, line 95: Nit: This should prevent default when handling `Enter`

Jomshir98 (changes requested):
I also don't agree with this solution for interaction subsystem... it feels easy to forget setting the flag, especially as more things get added. Wouldn't it be better to try exporting the data after loading it, and checking if they differ - and save then?

Jomshir98 (approved):
LGTM! (feel free to close the remaining discussion if you decide not to tackle it here)

Issue #265, opened 2023-06-02
Add module support for room device assets


ClaudiaMia:
Room device asset definitions currently do not support adding toggleable modules and lock modules like normal assets do.

PR #440, opened 2023-11-03
[FEATURE] Room device modules

Jomshir98:
Resolves: #265 

This PR adds module support to the room devices, namely:
- Modules are generic w.r.t. the "properties" they result in (this needed changes to the config, asset repo PR ports old assets)
- Reworks appearance actions to work more directly rather than through serializing and deserializing items every step of the way (was needed to preserve new links on room device wearable parts)
- Updates modules usage to work on both wearable part and room devices
- Adds a new kind of properties for room devices `RoomDeviceProperties`
- Adds support for image overrides to the static layers of room devices (+condition evaluator updates to make that happen)

Todo:
- Test things that worked before to find what was broken
- Make some room devices that use modules
- Test the new room devices
- Fix all that is broken


Sekkmer (approved):
looks good

(this could be done later or in assets repo: conditions could be separated/validated to be module type only on devices)

Issue #68, opened 2022-11-14
[META] Mobile improvements


Jomshir98:
# This issue is for tracking small mobile issues in general
It should be edited or commented on, but not closed.
Game-breaking issues should have separate issue created.
---

- [ ] Top menu should turn into hamburger menu on narrow screens, sliding onto screen only when opened
- [ ] DM menu should hide selection when friend is selected
- [ ] Settings menu has too large min width on fieldsets, causing unnecessary horizontal overflow
- [ ] Settings menu should separate horizontal scrolling of tabs and content
- [ ] Character selection screen is not centered
- [ ] Wardrobe: Cannot scroll when touching canvas even tho it has no interactivity
- [ ] Wardrobe: It is not mobile friendly at all in general

PR #443, opened 2023-11-08
[FEATURE] Collapsable header

Jomshir98:
Ref: #68 
Adds ability to collapse the header buttons into "hamburger" menu when the screen is narrow (e.g. on phones in portrait mode).
Also attempts to fix pinch issues with graphics viewport.

### DIFF
--- a/pandora-client-web/src/components/header/Header.tsx
+++ b/pandora-client-web/src/components/header/Header.tsx
@@ -24,7 +24,7 @@ import { LeaveButton } from './leaveButton';
 import { useIsNarrowScreen } from '../../styles/mediaQueries';
 import { Column, Row } from '../common/container/container';
 import { DialogInPortal } from '../dialog/dialog';
-import { Button } from '../common/button/button';
+import { IconCross, IconHamburger } from '../common/button/domIcons';
 
 function LeftHeader(): ReactElement {
 	const connectionInfo = useShardConnectionInfo();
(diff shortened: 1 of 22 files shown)


pandora-client-web/src/components/header/Header.tsx
```diff
@@ -146,15 +196,69 @@ function FriendsHeaderButton(): ReactElement {
...
+	const [showMenu, setShowMenu] = useState(false);
+
+	return (
+		<Row alignX='space-between' alignY='center' className='flex-1'>
+			<button className='HeaderButton' onClick={ () => {
+				setShowMenu(true);
+			} }>
+				<IconHamburger />
+			</button>
+			<span>{ currentAccount != null ? currentAccount.username : '[not logged in]' }</span>
```
ClaudiaMia, line 241: Thinking about it more, I feel it makes not much sense to show the acocunt name here. It is more likely you use different characters of the same account in different tabs, needing a way to quickly see which character the tab is.
Of course you can still show "no character selected" or "not logged in" there, too.

For consistency reasons, I would also recommend to switch the location of the hamburger menu to the right and the character name to the left, which doubles as a character menu, like in the big desktop version.

For the open hamburger menu, the X would then also need to travel to the right side.

ClaudiaMia (changes requested):

pandora-client-web/src/components/header/Header.tsx
ClaudiaMia: Got a live demo of my requested changes. I am now happy with it!~

ClaudiaMia (approved):

PR #451, opened 2023-11-27
[FIX] Add Wiki tests and downgrade react-router to avoid latest regression for nested routing

Jomshir98:
Tracking issue for upstream: https://github.com/remix-run/react-router/issues/11052


ClaudiaMia (approved):
LGTM!

PR #459, opened 2023-12-03
[FEATURE] preview cutter

Sekkmer:
- add `preview?: string | null` to asset definition (allow undefined until all assets have this property defined)
 - add preview cutter for the result layer (Preview tab), and controls on Items tab

### DIFF
--- a/pandora-client-web/src/components/wardrobe/wardrobeComponents.tsx
+++ b/pandora-client-web/src/components/wardrobe/wardrobeComponents.tsx
@@ -16,7 +16,7 @@ import { useWardrobeContext, useWardrobeExecuteChecked } from './wardrobeContext
 import { useStaggeredAppearanceActionResult } from './wardrobeCheckQueue';
 import _ from 'lodash';
 import { usePermissionCheck } from '../gameContext/permissionCheckProvider';
-import { useCurrentAccount } from '../gameContext/directoryConnectorContextProvider';
+import { useCurrentAccountSettings } from '../gameContext/directoryConnectorContextProvider';
 
 export function ActionWarningContent({ problems }: { problems: readonly AppearanceActionProblem[]; }): ReactElement {
 	const assetManager = useAssetManager();
@@ -228,12 +228,10 @@ export function InventoryAssetPreview({ asset, small }: {
 }
 
 function useAssetPreviewType(small: boolean): 'icon' | 'image' {
-	const account = useCurrentAccount();
-	if (!account)
-		return 'icon';
+	const settings = useCurrentAccountSettings();
 
 	if (small)
-		return account.settings.wardrobeSmallPreview;
+		return settings.wardrobeSmallPreview;
 
-	return account.settings.wardrobeBigPreview;
+	return settings.wardrobeBigPreview;
 }
--- a/pandora-client-web/src/components/wardrobe/wardrobeComponents.tsx
+++ b/pandora-client-web/src/components/wardrobe/wardrobeComponents.tsx
@@ -185,7 +185,10 @@ export function InventoryAssetPreview({ asset }: {
 }): ReactElement {
 	const assetManager = useAssetManager();
 
-	const iconAttribute = useMemo(() => {
+	const preview = useMemo(() => {
+		if (asset.definition.preview != null)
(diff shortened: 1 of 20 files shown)


pandora-client-web/src/editor/components/previewCutter/previewCutter.tsx
```diff
@@ -0,0 +1,221 @@
...
+		return () => {
+			editorSceneContext = null;
+		};
+	}, [context]);
+
+	return (
+		<Graphics
+			zIndex={ 0 }
+			interactive={ true }
+			hitArea={ new PIXI.Rectangle(x, y - lineWidth / 2, size, size) }
```
Jomshir98, line 122: Why is `y` shifted by `lineWidth / 2`, but `x` isn't? (actually I think neither should be)

pandora-client-web/src/editor/components/previewCutter/previewCutter.tsx
```diff
@@ -0,0 +1,221 @@
...
+				<label htmlFor='preview-cutter-x'>X</label>
+				<input id='preview-cutter-x' type='number' value={ state.position.x } onChange={ setX } />
+			</div>
+			<div>
+				<label htmlFor='preview-cutter-y'>Y</label>
+				<input id='preview-cutter-y' type='number' value={ state.position.y } onChange={ setY } />
+			</div>
+			<div>
+				<label htmlFor='preview-cutter-size'>Size</label>
+				<input id='preview-cutter-size' type='number' value={ state.size } onChange={ setSize } />
```
Jomshir98, line 208: This one should use either `useRemotelyUpdatedUserInput` or the following suggestion to be more user friendly.
The reason is, that it cannot save all values users might enter, which causes problems when one tries to enter a new value.
For example consider following case:
- Current value is `250`.
- I want to enter `125`, so I select the whole number and start typing `1`
- At this moment the value changes to `50` (`PREVIEW_CUTTER_MIN_SIZE`).
```suggestion
				<input id='preview-cutter-size' type='number' defaultValue={ state.size } onChange={ setSize } />
```

Jomshir98 (changes requested):
Overall I like it. A few minor notes to the preview cutter itself, and one bigger missing feature that I think should be done as part of this PR before it is merged.

pandora-client-web/src/editor/components/previewCutter/previewCutter.tsx
```diff
@@ -0,0 +1,221 @@
...
+import { Button } from '../../../components/common/button/button';
+import { useEvent } from '../../../common/useEvent';
+import { ImageExporter } from '../../graphics/export/imageExporter';
+import { DownloadAsFile } from '../../../common/downloadHelper';
+import { EditorSceneContext, useEditorSceneContext } from '../../graphics/editorScene';
+import './previewCutter.scss';
+
+type PreviewCutterState = Readonly<{
+	enabled: boolean;
+	lineWidth: number;
```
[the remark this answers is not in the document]
Sekkmer: I had this for testing since it is more visible what is wrong with a higher lineWidth, removed it and moved it to a const

pandora-client-web/src/editor/components/previewCutter/previewCutter.tsx
Sekkmer: added a new `InputNumber` element that should handle this much better

pandora-client-web/src/components/wardrobe/wardrobeComponents.tsx
```diff
@@ -197,12 +200,12 @@ export function InventoryAssetPreview({ asset }: {
...
 		if (filterAttribute)
-			return filterAttribute;
+			return filterAttribute.icon;
 
-		return validAttributes.length > 0 ? validAttributes[0] : undefined;
+		return validAttributes.length > 0 ? validAttributes[0].icon : undefined;
 	}, [asset, assetManager]);
 
-	const icon = useGraphicsUrl(iconAttribute?.icon);
+	const icon = useGraphicsUrl(preview);
```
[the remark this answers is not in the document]
Sekkmer: added the 2 options

Jomshir98 (approved):

Issue #12, opened 2022-08-29
Character / Account view


Jomshir98:
It should be possible to display information screen for character, including at least following info:
- Name and ID of character
- Description of character - which can be opened in subscreen and edited if opened by Player
- Visible name and ID of account
- Age of account (since creation)
- Account Description - which can be opened in subscreen and edited if opened by Player
- Place to display relationships (owner, lovers, possibly submissives) - not implemented at this stage, but with place reserved for it

**Note, that focus of this issue is only to prepare the view graphically, not implement the descriptions loading and saving**; see #11

Issue #11, opened 2022-08-29
Character / Account description


Jomshir98:
Both character and account have independent descriptions
- Account description is visible from all characters
- Both are visible by people inside room

Jomshir98:
Depends on #12 to have a screen for showing this data.

PR #461, opened 2023-12-04
[ADD] Account and character profiles

Jomshir98:
Resolves: #11
Resolves: #12 

Adds a new screen for showing information about an account or a character.
Information about character can be seen for characters in the same room.
Information about accounts can be seen for characters in the same room and friends (or by being a moderator).

Also adds description to both accounts and characters.


pandora-server-directory/src/account/accountRelationship.ts
```diff
@@ -99,6 +99,48 @@ export class AccountRelationship {
 		return false;
 	}
 
+	/**
+	 * Checks if this account's profile is visible to the passed account
+	 * @param queryingAccount - The account attempting to access this account's profile
+	 * @returns - If the access should be allowed
+	 */
+	public async profileVisibleTo(queryingAccount: Account): Promise<boolean> {
```
Sekkmer, line 108: this is really similar to `canReceiveDM(from: Account)` function

could we merge it somehow, only difference is the moderator part and the allowDirectMessagesFrom, which I feel wouldn't be an issue if profile would be visible to all in that case

dms also have username, maybe we could use displayName there too

pandora-server-directory/src/account/accountRelationship.ts
```diff
@@ -99,6 +99,48 @@ export class AccountRelationship {
...
+		// Player can always see their own profile
+		if (this.account.id === queryingAccount.id)
+			return true;
+
+		// Moderators can see anyone's profile
+		if (queryingAccount.roles.isAuthorized('moderator'))
```
Sekkmer, line 114: unrelated to this PR, just an idea

maybe we could add a state flag to the account that toggles moderator functionality \
so if you just play normally, you would still be affected by blocks and such, but \
if you have a moderator role, you can enable settings that let you bypass such limitations

pandora-server-directory/src/account/accountRelationship.ts
Jomshir98: I simplified the "characters share a room" bit by pulling it out into separate function. Otherwise I don't feel the similarity is a problem.
About just letting anyone see - I would prefer not to, but in the end it is just a preference.

For `displayName` - currently it is not configurable and just uses username;
I plan to go through existing code after this PR is merged and replace all cases where username is exposed directly with fully configurable display name. I'm not even sure yet if it should default to the username or to something else, e.g. `User <id>`.

pandora-server-directory/src/account/accountRelationship.ts
Jomshir98: I'm thinking about simply moving the moderator bits into the existing `Management` menu and putting any moderation tools there. I don't think it is a good idea to allow moderators to have any extra "IC" powers at all - only account-based ones.
If people get stuck we don't want to help them anyway - what would then be point of Pandora? And other cases where moderator would interfere with existing tools I think are all about (at least partially) banning someone.

Sekkmer: What I meant is to only allow moderators to see user profiles to which they have no access otherwise if they truly wish to do this as a moderator, so a limitation to this exact functionality, not any "IC" powers

(feel free to close this conversation)

Sekkmer (approved):

PR #464, opened 2023-12-09
[ADD] State flag support for room devices

Jomshir98:
Adds same state flags support as normal items have to room device modules - allowing room devices to limit which internal states are valid.


ClaudiaMia (approved):
LGTM

PR #466, opened 2023-12-10
[CHANGE] improvements based on Steffi's feedback

ClaudiaMia:
Primarily changed the user facing parts of the relationship feature to "contacts" since the name was not fitting and we also later on have a real relationship feature planned. Also changed the user facing mentioning of the word "friend", as it was misleading.

Introduced ways for users to easily see if a room is public or private.


pandora-client-web/src/components/chatroomSelect/chatroomSelect.tsx
```diff
@@ -25,9 +26,10 @@ const TIPS: readonly string[] = [
 	`Your character can turn around for everyone in a chat room in the "Pose" tab or with "/turn".`,
 	`Chat commands start with a "/" and typing just this one character shows a help menu.`,
 	`You can use your browser's "back" and "forward" buttons to navigate between screens.`,
-	`The dragging points to move room devices are invisible, but generally under the item.`,
 	`In the Pandora settings, character (chat) and account (direct messages) name colors are set separately.`,
 	`Every single change in the wardrobe happens instantly and is immediately visible to everyone in the room.`,
+	`You need to equip the handheld item "room construction tools" to decorate your rooms`,
```
Sekkmer, line 31: `to decorate your rooms` could be changed to `to setup devices` or something similar that is a bit more clear

pandora-client-web/src/components/settings/accountSettings.tsx
```diff
@@ -10,15 +10,26 @@ import _, { uniq } from 'lodash';
...
 export function AccountSettings(): ReactElement | null {
+	const navigate = useNavigate();
 	const account = useCurrentAccount();
 
 	if (!account)
 		return <>Not logged in</>;
 
 	return (
 		<>
+			<Button className='slim' onClick={ () => { // temporary button
```
Sekkmer, line 24: why temporary, when will it be removed/changed?

also it is probably better to add `TODO` in comments like these so we can search for them

pandora-client-web/src/components/releationships/relationships.tsx
```diff
@@ -25,10 +25,10 @@ export function Relationships() {
 	return (
 		<div className='relationships'>
 			<UrlTabContainer>
-				<UrlTab name={ <RelationshipHeader type='friend' /> } urlChunk='friends'>
+				<UrlTab name={ <RelationshipHeader type='friend' /> } urlChunk='list'>
```
Sekkmer, line 28: shouldn't this change be: `frients` -> `contacts`? instead of the `list`

Sekkmer (changes requested):

pandora-client-web/src/components/releationships/relationships.tsx
ClaudiaMia: Well, the URL would be `contacts/contacts` then, which is a bit silly. It could also be changed to `contacts/main` maybe.

pandora-client-web/src/components/settings/accountSettings.tsx
ClaudiaMia: Good point, but I am not yet fully sure what to do with them in the future. Will discuss it with Jomshir, too.

Sekkmer (approved):

Jomshir98 (approved):

PR #475, opened 2023-12-14
[CHANGE] wardrobe preview style

ClaudiaMia:


pandora-client-web/src/components/wardrobe/wardrobe.scss
```diff
@@ -265,8 +265,15 @@ $drop-overlay-gap: 0.3em;
...
-		grid-template-columns: repeat(3, minmax(0, 384px));
 		grid-template-rows: min-content auto;
+
+		&.small {
+			grid-template-columns: repeat(3, minmax(0, 384px));
+		}
+
+		&.large {
+			grid-template-columns: repeat(4, minmax(0, 384px));
+		}
```
Sekkmer, line 276: why not just do a media quey instead?

Sekkmer (changes requested):

pandora-client-web/src/components/wardrobe/wardrobe.scss
ClaudiaMia: Because.... I did not even know that something like that exists. *giggles*
I looked it up just now and committed a change. I like it, as it now changes the layout dynamically on window changes and not only on react state changes. Thank you!~

Sekkmer (approved):

Jomshir98 (approved):

PR #484, opened 2023-12-17
[FIX] Fixes and improvements for graphics suspense and editor

Jomshir98:
Fixes several crashes in editor.
Improves graphics suspense:
- Both container updates and loading progress updates happen synchronously and outside React now
- This avoids previous error in editor / when using custom texture getter
- Adds a grace period before showing progress again after being ready


ClaudiaMia (approved):
Tested it locally and the bugs are gone. Everything works as expected now. Thank you!

PR #486, opened 2023-12-18
[FEATURE] asset preferences for characters

Sekkmer:
Types for assets:
 - Favorite
 - Normal
 - Maybe
 - Prevent
 - Do not render

Types for attributes:
 - Normal
 - Prevent
 - Do not render

Progress:
 - [x] sort wardrobe view based on preference
 - [x] show "start" for favorite
 - [x] grayscale maybe/prevent/do not render
 - [x] GUI to set asset preferences
 - [x] GUI to set attribute preferences
 - [x] only allow setting "Do not render" when asset is not worn
 - [x] skip rendering assets with "Do not render" preference
 
 Parts that won't be included:
  - device preferences
  - character / role based overrides
  - account wide settings to configure UI based on preferences


pandora-server-directory/src/database/dbHelper.ts
```diff
@@ -23,6 +23,7 @@ export function CreateCharacter<Id extends number | CharacterId>(accountId: numb
 		created: -1,
 		accessId: nanoid(8),
 		settings: cloneDeep(CHARACTER_DEFAULT_PUBLIC_SETTINGS),
+		preferences: cloneDeep(ASSET_PREFERENCES_DEFAULT),
```
ClaudiaMia, line 26: Missing import prevents testing this feature.

ClaudiaMia:
Some things are likely not yet implemented, as it is a draft, but I played around a bit with it and this is what I observed:

- You yourself can use a an item on yourself that is "prevented from being used". I feel this is correct and what I would expect.
- Others can correctly not use the item that is "prevented from being used" on yourself. But I wonder when this is synced. Changing the preference back to normal permanently prevents my other test account from adding the item, so far as I tested opening and closing wardrobes and waiting a while.

About "grayscale maybe/prevent/do not render, (hide?, account wide config?)":
I feel that it should indeed be account wide and it should not render such items on yourself from your own perspective as well as on other characters from your own perspective, but still show the assets in the inventory or as worn.

ClaudiaMia (commented):
Some things we noticed:

WE do not think the favorite or any other state should affect sorting at all! Instead we propose highlighting:

For the player:
favorite: add a yellow star symbol (e.g. after the asset name ? - maybe highlight the item name too somehow )
normal: normal background color
maybe: light orange background color
prevent: dark red background color (might need switching text color to black, or make it a light red only)
not render: same red as prevent with preview forced to something like an "X"-image, including the image-based preview.

For other players opening your wardrobe:
favorite: add a yellow star symbol (after the name ? - maybe highlight the item name too somehow )
normal: normal background color of the row or tile
maybe: light orange background color
prevent AND not render: dark red background color (might need switching text color to black, or make it a light red only)

Some more findings:
- rows are buttons for some reason, even though clicking them does nothing
-  some [?] icons are no longer square, they should force-set width to correct one too, if possible
- A wiki page is definitely needed to explain the details - I would opt for a full page instead of just another paragraph
  - Maybe naming the wiki page more generally - "Limits"? And only header so far would be "Limiting worn assets"
- Typo in the hover text:
Do not render this items with this attribute. -> Do not render items with this attribute

ClaudiaMia (changes requested):
Some more issues I found:

(1) When switching Settings > Interface > Wardrobe UI > list mode: show preview images, all list views in the wardrobe no longer limit the size of the images, and they are huge.

(2) It feels a bit weird to allow all those attributes without an icon, as this might be confusing for users and it is also a lot of text to read through to find something specific. So while this offers great detailed mass-blocking, I feel individual item blocking is much better and clearer, as you know exactly what results you will get. Also, since assets can have many attributes, this complexity can lead to issues (see next issue) so I might even vote for removing attribute based limiting at all!

(3) If I set something as favorite, and then block an attribute this is contained in, the item stays a favorite. Now, this could be intended or not, but for the user the behavior is not obvious independent of which of the two results will happen.
 Also, if I then change the favorite asset back to normal and then the attribute also to normal and block the attribute again, the once manually set individual asset stays normal so there seems to be no way to reset assets again to be mass-changeable via the attribute system.
I see this as yet another proof that attribute based limiting is hard to make work intuitively and how the user may expect it and should rather be removed completely, as it could lead to unpleasant and limit breaking situation if it is not clear enough how it works.

ClaudiaMia:
If we would remove the attribute based preferences, then we could use the space on the left side to show the text on how this is used and what it does and means instead.

ClaudiaMia:
> Some more issues I found:
> 
> 
> (2) It feels a bit weird to allow all those attributes without an icon, as this might be confusing for users and it is also a lot of text to read through to find something specific. So while this offers great detailed mass-blocking, I feel individual item blocking is much better and clearer, as you know exactly what results you will get. Also, since assets can have many attributes, this complexity can lead to issues (see next issue) so I might even vote for removing attribute based limiting at all!
> 
> (3) If I set something as favorite, and then block an attribute this is contained in, the item stays a favorite. Now, this could be intended or not, but for the user the behavior is not obvious, independent of which of the two results will happen. Also, if I then change the favorite asset back to normal and then the attribute also to normal and block the attribute again, the once manually set individual asset stays normal so there seems to be no way to reset assets again to be mass-changeable via the attribute system. I see this as yet another proof that attribute based limiting is hard to make work intuitively and how the user may expect it and should rather be removed completely, as it could lead to unpleasant and limit breaking situations for users if it is not clear enough for them how it works.

Jomshir has some ideas on how to rework the UI in a way that let me change my opinion to remove asset based limiting completely.

pandora-common/src/assets/graphics/graphics.ts
```diff
@@ -1,5 +1,5 @@
 import { z } from 'zod';
-import type { AssetId } from '../definitions';
+import type { AssetId } from 'assets/base';
```
Jomshir98, line 2: This path should be relative.

pandora-common/src/character/characterData.ts
```diff
@@ -21,13 +22,41 @@ export const CHARACTER_DEFAULT_PUBLIC_SETTINGS: Readonly<ICharacterPublicSetting
...
+export const AssetPreferencesSchema = z.object({
+	attributes: z.record(z.string(), AttributePreferenceSchema).default({}),
+	assets: z.record(AssetIdSchema, AssetPreferenceSchema).default({}),
+});
+export type AssetPreferences = z.infer<typeof AssetPreferencesSchema>;
+
+export const ASSET_PREFERENCES_DEFAULT: Readonly<AssetPreferences> = Object.freeze<AssetPreferences>({
+	attributes: {},
+	assets: {},
+});
```
Jomshir98, line 50: Wouldn't it be worth putting these into a new file?

pandora-common/src/character/characterData.ts
```diff
@@ -21,13 +22,41 @@ export const CHARACTER_DEFAULT_PUBLIC_SETTINGS: Readonly<ICharacterPublicSetting
...
+
 /** Data about character, that is visible to everyone in same room */
 export const CharacterPublicDataSchema = z.object({
 	id: CharacterIdSchema,
 	accountId: AccountIdSchema,
 	name: CharacterNameSchema,
 	profileDescription: z.string().default(''),
 	settings: CharacterPublicSettingsSchema.default(CHARACTER_DEFAULT_PUBLIC_SETTINGS),
+	preferences: AssetPreferencesSchema.default(ASSET_PREFERENCES_DEFAULT),
```
Jomshir98, line 59: Naming it just `preferences` seems weird when it specifically only concerns asset preferences - maybe name it that - `assetPreferences`?

pandora-common/src/character/restrictionsManager.ts
```diff
@@ -540,3 +569,21 @@ export class CharacterRestrictionsManager {
 		return { allowed: true };
 	}
 }
+
+export function ResolveAssetPreference({ preferences }: Immutable<{ preferences: AssetPreferences; }>, asset: Asset, _source?: CharacterId): AssetPreferenceType {
```
Jomshir98, line 573: Maybe this could receive only the preferences instead of wrapping them in an object and then immediately destructing?

pandora-common/src/gameLogic/character/character.ts
```diff
@@ -15,14 +16,16 @@ export abstract class GameLogicCharacter extends TypedEventEmitter<GameLogicChar
 	public readonly id: CharacterId;
 	public readonly accountId: AccountId;
 	public readonly name: string;
+	public readonly publicData: Immutable<ICharacterPublicData>;
```
Jomshir98, line 19: This is problematic. Here you have a `readonly` property that is marked as `Immutable`, while in fact you expect it to be updated elsewhere. I really dislike such pattern. It also relies heavily on multiple places having reference to a same object, which is an invariant that is easy to break.

Instead you should look into how the `InteractionSubsystem` is done - this character really should only have the minimal data (which is special because it _never_ changes after character creation is finished).
The asset preferences should then be added as `abstract readonly` in this class, an updatable property in the `GameLogicCharacterClient` and either an updatable property or getter (a function passed through constructor) to the `GameLogicCharacterServer` class.

ClaudiaMia:
> Some more issues I found:
> 
> (1) When switching Settings > Interface > Wardrobe UI > list mode: show preview images, all list views in the wardrobe no longer limit the size of the images, and they are huge. This was introduced by this branch.

The last commit fixed it partially, but not fully as the images are now too small in certain row views, where they utilized the full available space in the row (stretching from top border to bottom border) on `master`.

We have two sizes of rows in the wardrobe, the normal one and the slim one used for instance in the create tab. Those two types need different small preview image sizes.

pandora-common/src/utility.ts
```diff
@@ -407,3 +407,17 @@ export function PromiseOnce<T>(promise: () => Promise<T>): () => Promise<T> {
 		return result[0];
 	};
 }
+
+export type ObjectEntry<T, K extends keyof T> = [K, T[K]];
+
+export class Obj {
```
Jomshir98, line 413:
```suggestion
/**
 * This class provides alternative implementations of `Object.keys`, `Object.values` and `Object.entries`,
 * which have typing as if all possible keys of the object passed to them are known.
 * This results in a more friendly way to iterate through objects where that is the case,
 * but it can still result in an **incorrect behaviour** if you don't know what exactly this does.
 *
 * Use with care.
 */
export class KnownObject {
```

pandora-common/src/assets/definitions.ts
```diff
@@ -473,6 +471,8 @@ export type AssetAttributeDefinition<A extends AssetDefinitionExtraArgs = AssetD
 		tab: 'item' | 'body' | 'room';
 		excludeAttributes?: readonly A['attributes'][];
 	};
+	/** Do not allow this attribute to be used as a preference */
+	noPreference?: true;
```
Jomshir98, line 475: Suggested type:
```
/** Set to false to prevent this attribute from being used to specify asset preferences/permissions/blocks */
useAsAssetPreference?: false;
```

Jomshir98 (changes requested):
This is a review of only `pandora-common` changes so far. Once these are resolved, I will look at the rest.

pandora-common/src/gameLogic/character/character.ts
Sekkmer: other then this I fixed the rest, will look at this tomorrow

Sekkmer: well I looked into this, but wherever I insert a getter and or a subscription, it just feels like pure bloat that does nothing but add more code with extra complexity and more things to do for the server while being completely useless

if we somehow need to track the event when the change occurs then I can see this as justifiable but as it currently stands I do not see this as a potential improvement

`Immutable` is nothing more than a deep readonly, it doesn't mean a persistent or frozen object, in this case, it is just a constant reference

JS references are the simplest way to have a shared state, and the Immutable ensures that from `GameLogicCharacter` cannot modify it, I would not throw out the simplicity of using a reference unless using reference is not possible

we can just rename the variable to `publicDataRef` or something and be done with it

Jomshir98 (approved):
Looks good now! I did a few more edits while reviewing this instead of suggesting them, the commits contain my reasoning. Feel free to poke me if you disagree with any.

ClaudiaMia (approved):
Good to merge for further UI rework in a new PR.

Sekkmer:
ehh why can't GH be consistent about `--committer-date-is-author-date`...

PR #487, opened 2023-12-18
[FEATURE] Character's personal room

Jomshir98:
Works towards: #70 

This PR adds personal rooms to characters.

All personal room's data is saved on the character itself.
Shard and client have been modified to think about a character as always being in _some_ room.
The personal room cannot be entered by anyone but the character.

Currently is in a rather unstable state - while major work is done, I'm keeping it as a draft until it is more thoroughly tested.
TODO:
- [x] Test, test, test and test
- [x] Add information to the wiki
- [x] Add context help to the personal room's interface


pandora-client-web/src/components/chatroomSelect/chatroomSelect.tsx
ClaudiaMia, line 1: Line 66: "< Back to lobby" should be renamed to just "< Back"

Also do we need the file `pandoraLobby.tsx` still?

Sekkmer (approved):

PR #498, opened 2023-12-30
[CHANGE] Item limits UI improvements

Jomshir98:
Depends on a pandora-assets PR to update attribute definitions.

This PR reworks most of the item limits (asset preferences) UI and adds few related changes:
- The asset preferences UI has been reworked
- Improved validation message about limited items/attributes
- Allowed "Maybe" for attributes
- Distinguished between item not having anything chosen (is based on attributes) and being set to normal (even if all attributes are "normal" too)
- Changed the resolution behavior not to use Player's DoNotRender setting - instead the Player's DoNotRender is only applied to the preview - the name and icon represent target's preference.
- Improved how attribute filter buttons work and where they show


pandora-client-web/src/components/common/select/select.tsx
```diff
@@ -6,9 +6,15 @@ export interface SelectProps extends Omit<DetailedHTMLProps<SelectHTMLAttributes
 		currentTarget: HTMLSelectElement;
 		target: HTMLSelectElement;
 	}): void;
+	noScrollChange?: boolean;
 }
 
-export const Select = React.forwardRef<HTMLSelectElement, SelectProps>(function Select({ children, onChange, ...props }, ref): ReactElement {
+export const Select = React.forwardRef<HTMLSelectElement, SelectProps>(function Select({
```
Sekkmer, line 12: I was considering this the last time I made the PR but eventually, I decided against it, this property basically makes this whole element do nothing and work as a simple `<select>` element

we do not use this property dynamically so I see no benefit from adding this over simply just using `<select>`

pandora-common/src/character/assetPreferences.ts
```diff
@@ -23,8 +23,8 @@ export const AssetPreferenceSchema = z.object({
 export type AssetPreference = z.infer<typeof AssetPreferenceSchema>;
 
 export const AssetPreferencesSchema = z.object({
-	attributes: z.record(z.string(), AttributePreferenceSchema).default({}),
-	assets: z.record(AssetIdSchema, AssetPreferenceSchema).default({}),
+	attributes: z.record(z.string(), AttributePreferenceSchema.optional()).default({}),
```
Sekkmer, line 26: why would we want to allow optional here, `undefined` is removed from JSON anyway, and we certainly do not want to store `undefinied` in database, so I think the validation should not accept if a key exists with `undefined` as a value

pandora-common/src/character/assetPreferences.ts
```diff
@@ -129,20 +131,6 @@ export function CleanupAssetPreferences(assetManager: AssetManager, {
...
-					hasAttribute = true;
-					break;
-				}
-			}
-			if (!hasAttribute) {
-				delete assets[key];
-				hasInvalid = true;
-				continue;
-			}
-		}
```
Sekkmer, line 121: I am not sure if it is a good idea to remove this, assets that do not have any attribute affecting them and are in default state have no effect and should be cleaned if possible

The only reason to allow them is to be user-friendly so you can put a currently worn item in "normal" and then change an attribute to "doNotRender" which would not be possible otherwise

and the same for attributes unless we extend the base option with other thing "normal" should be dropped and not stored in database, ideally we should try to stay as sparse as we can

Sekkmer (changes requested):

pandora-common/src/character/assetPreferences.ts
Jomshir98: Because then TS correctly marks value gotten from the record as potentially `| undefined`.
It also allows doing `{ ...oldValue, [id]: undefined }`, which works well, but if passed to Zod (which happens during send too), it screams about it.

pandora-client-web/src/components/common/select/select.tsx
Jomshir98: I agree, but I'm looking forward to the future: It is not unlikely we will add more features to the `Select` element, like being able to work with non-trivial values. For that case we want to reuse it in as many places as possible to be consistent. Also it gives a precedent for people that copy existing code to make new one.

pandora-common/src/character/assetPreferences.ts
Jomshir98: The reason is to preserve the item's setting - not changing unexpectedly.
Imagine the following flow:
- User sets some attribute to "Maybe"
- One items blocked by it is fully okay for them, however, so they override that item to "Normal"
- *) Now they go the attribute, wanting to change it to "Prevent", but they accidentally (or temporarily intentionally) change it to "Normal"
- Now change it correctly to "Prevent"
- ... but the item is now set as "Prevent", despite them having set it manually to "Normal"

To make things even more strange this behaviour is not consistent if you insert server restart/shard switch after the `*)` point.
Also instead of this example of `*)` point you can imagine asset definition changing to not have that attribute and later the attribute being re-added - which is even worse as user's preference was lost for them seemingly randomly.

pandora-common/src/character/assetPreferences.ts
Jomshir98: Hmm~ But I guess it is not that difficult to do it differently...

Sekkmer: the first is unfortunate, I do not like the things around this JS/TS behavior issue
for the second we could just use `omit({ ...oldValue }, id)`

I guess that change is fine, I just do not like it

pandora-common/src/character/assetPreferences.ts
Jomshir98: Oh, and for attributes: For attributes I drop it always (both in transit and in database) - I did that by changing the update call to contain full data, so lack of entry means resetting it to the default value.

Sekkmer: Well, I know this cleanup can cause slightly unexpected behavior.

I would ignore assets repo changes since that can be viewed as unexpected regardless since the user may not be aware what attribute changed for items.

So, with that out of the way, step 3 only causes issues if and only if the server restarts, which I do not see as that big of a problem, though delaying it more could be an option to only happen at a fixed time after a change has been made (maybe a week since last change, the user surely won't remember any of it)

But if you feel this data is not worth the hassle to cleanup and should not effect effect size in a meaningful way feel free to close this thread

Sekkmer (approved):

PR #499, opened 2023-12-30
[CHANGE] device deployment improvements

### LINKED ISSUE #455: Persistent room device coordinates
When I store a room device, I want the current room coordinates to be preserved while the room device is in the room inventory, so deploying it again puts it on the last spot.

In case the room background was changed in the meantime and that position is illegal, it would just fall back to the default 0,0.

### LINKED ISSUE #494: Add custom warning message if a device is occupied and the user want to store it

Sekkmer:
resolves #494
resolves #455 (position is preserved even if background changes)

(currently deployed devices should stay deployed after the change)

### DIFF
--- a/pandora-client-web/src/components/chatroom/contextMenus/deviceContextMenu.tsx
+++ b/pandora-client-web/src/components/chatroom/contextMenus/deviceContextMenu.tsx
@@ -16,6 +16,7 @@ import { IChatRoomMode } from '../chatRoomScene';
 import { toast } from 'react-toastify';
 import { ActionWarningContent } from '../../wardrobe/wardrobeComponents';
 import { TOAST_OPTIONS_WARNING } from '../../../persistentToast';
+import { omit } from 'lodash';
 
 function StoreDeviceMenu({ device, close }: {
 	device: ItemRoomDevice;
@@ -53,7 +54,7 @@ function MoveDeviceMenu({ device, setChatRoomMode, close }: {
 			itemId: device.id,
 		},
 		target: { type: 'roomInventory' },
-		deployment: { deployed: true, position: { x: 0, y: 0, yOffset: 0 } },
+		deployment: { deployed: true, position: omit(device.deployment, 'deployed') },
 	}), [device]);
 	const checkResult = useStaggeredAppearanceActionResult(action, { immediate: true });
 	const available = checkResult != null && checkResult.problems.length === 0;
--- a/pandora-client-web/src/components/chatroom/contextMenus/deviceContextMenu.tsx
+++ b/pandora-client-web/src/components/chatroom/contextMenus/deviceContextMenu.tsx
@@ -1,5 +1,5 @@
 import { nanoid } from 'nanoid';
-import { ItemRoomDevice, AppearanceAction, ItemId, ICharacterRoomData, CloneDeepMutable } from 'pandora-common';
+import { ItemRoomDevice, AppearanceAction, ItemId, ICharacterRoomData } from 'pandora-common';
 import React, { useMemo, useState, ReactElement, useEffect, useCallback } from 'react';
 import { Character, ICharacter, useCharacterData } from '../../../character/character';
 import { ChildrenProps } from '../../../common/reactTypes';
@@ -28,7 +28,7 @@ function StoreDeviceMenu({ device, close }: {
 			itemId: device.id,
 		},
 		target: { type: 'roomInventory' },
-		deployment: null,
+		deployment: { deployed: false },
 	}), [device]);
 	const checkResult = useStaggeredAppearanceActionResult(action, { immediate: true });
 	const available = checkResult != null && checkResult.problems.length === 0;
@@ -53,7 +53,7 @@ function MoveDeviceMenu({ device, setChatRoomMode, close }: {
 			itemId: device.id,
 		},
 		target: { type: 'roomInventory' },
-		deployment: CloneDeepMutable(device.deployment),
+		deployment: { deployed: true, position: { x: 0, y: 0, yOffset: 0 } },
 	}), [device]);
 	const checkResult = useStaggeredAppearanceActionResult(action, { immediate: true });
 	const available = checkResult != null && checkResult.problems.length === 0;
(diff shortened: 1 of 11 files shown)


ClaudiaMia (changes requested):
Tested and the room device positioning works great and as expected.

However, it does not seem to fix https://github.com/Project-Pandora-Game/pandora/issues/494
when trying to deploy an occupied room device either in the context menu or via the button when having it selected in the room inventory, in both cases I got the old generic toast message.

pandora-common/src/assets/appearanceActions.ts
```diff
@@ -945,13 +945,25 @@ export function ActionRoomDeviceDeploy(processingContext: AppearanceActionProces
 	if (!manipulator.modifyItem(itemId, (it) => {
 		if (!it.isType('roomDevice'))
 			return null;
+
+		if (deployment === null && it.slotOccupancy.size > 0) {
```
Jomshir98, line 949: `deployment` can never be `null`, this likely wants to check that it has `deployed: false`?

pandora-common/src/assets/appearanceActionProcessingContext.ts
```diff
@@ -151,47 +165,25 @@ export class AppearanceActionProcessingResultInvalid extends AppearanceActionPro
...
 		super(processingContext);
-		this.problems = [
-			...processingContext.actionProblems,
-			{
-				result: 'invalidAction',
-				reason: invalidReason,
-			},
-		];
+		this.problems = [...processingContext.actionProblems];
 	}
```
Jomshir98, line 171: In that case here should be an assert that `problems` length is `> 0`

pandora-common/src/assets/appearanceActionProcessingContext.ts
```diff
@@ -125,10 +125,24 @@ export class AppearanceActionProcessingContext {
...
+		if (!validationResult.success) {
+			this.addProblem({
+				result: 'validationError',
+				validationError: validationResult.error,
+			});
+		}
+		if (this._actionProblems.length > 0)
+			return new AppearanceActionProcessingResultInvalid(this);
+
 		return new AppearanceActionProcessingResultValid(this, this.manipulator.currentState);
```
Jomshir98, line 146: This change changes the behaviour of validation quite a lot.
The new behaviour means, that if there is any problem at all, the resulting state is not made available (even if the state itself is technically valid, just the transition isn't, e.g. due to blocked hands or permissions).
This removes the possibility of doing previews for actions that you can't do right now but are otherwise valid.
I'm not fully sure which way I prefer more, but I want to be really sure you intended to do this change.

pandora-common/src/assets/appearanceActions.ts
```diff
@@ -945,13 +945,25 @@ export function ActionRoomDeviceDeploy(processingContext: AppearanceActionProces
...
+		}
+
 		previousDeviceState = it;
 		return it.changeDeployment(deployment);
 	}))
 		return false;
 
 	// Change message to chat
-	if (previousDeviceState != null && (deployment == null) !== (previousDeviceState.deployment == null)) {
+	if (previousDeviceState != null && (deployment == null) !== (!previousDeviceState.isDeployed())) {
```
Jomshir98, line 966: Same with `deployment == null` here

pandora-client-web/src/components/chatroom/contextMenus/deviceContextMenu.tsx
```diff
@@ -53,7 +53,7 @@ function MoveDeviceMenu({ device, setChatRoomMode, close }: {
 			itemId: device.id,
 		},
 		target: { type: 'roomInventory' },
-		deployment: CloneDeepMutable(device.deployment),
+		deployment: { deployed: true, position: { x: 0, y: 0, yOffset: 0 } },
```
Jomshir98, line 56: Nitpick: Here I would recommend using the existing position to check - the idea is that existing position must already be valid, so we are unlikely to hit some banned value. (if possible _values_ change in the future someone could easily forget the constants in here)

Jomshir98 (changes requested):

ClaudiaMia (changes requested):
I would like to request one more change as part of this feature:

Currently, when you try to store a deployed room device, there is a confirmation popup that says:
_**You might not be able to undo this action easily**
Storing a room device removes all characters from it
Storing a room device resets its position inside the room_

With your change this popup can be complete removed I think, as you can no longer undeploy while a device is occupied and the position is also no longer lost. Moreover, you cannot remove them anyway unless you are explicitly in the construction mode.

pandora-common/src/assets/appearanceActionProcessingContext.ts
Sekkmer: hmm, I made this change since I was debugging where the device store got invalid and it was super confusing to get a`ResultValid` class when the result was shown as not valid in the end

Sekkmer:
> I would like to request one more change as part of this feature:
> 
> Currently, when you try to store a deployed room device, there is a confirmation popup that says: _**You might not be able to undo this action easily** Storing a room device removes all characters from it Storing a room device resets its position inside the room_
> 
> While this confirmation popup is likely still needed for cases where entering a device would get yourself stuck with no way out, I am not sure such a confirmation dialog still needs to be shown when you merely try to store some a room device. The reason is that you introduce changes that you can no longer undeploy while a device is occupied and the position is also no longer lost. Moreover, you cannot remove them anyway unless you are explicitly in the construction mode.

warnings about device store has been removed in [this commit](https://github.com/Project-Pandora-Game/pandora/pull/499/commits/e4033f0205a02d5bed708b9e711d6956da277f4e)

pandora-common/src/assets/appearanceActions.ts
Sekkmer: yep and this should have fixed the issue Claudia mentioned above

pandora-common/src/assets/appearanceActionProcessingContext.ts
Sekkmer: I feel fine with not being able to show such previews, player can not get that state anyway so why show the preview

ClaudiaMia (approved):
Tested the changes and can confirm that the two things I reported are now working.

Jomshir98 (approved):

PR #500, opened 2023-12-30
[FIX] password locks should be case insensitive

### LINKED ISSUE #332: Password locks are not case insensitive
In the documentation it is written that password locks should behave case-insensitively, but the actual behaviour is case-sensitive.

The problem is most likely on this line, where the passwords are checked exactly, not case-insensitively: https://github.com/Project-Pandora-Game/pandora/blob/2a4f856782b0704669e4c1e3194171efbbbad30e/pandora-common/src/assets/item.ts#L852

Sekkmer:
fixes #332

### DIFF
--- a/pandora-common/src/assets/item.ts
+++ b/pandora-common/src/assets/item.ts
@@ -1177,7 +1177,8 @@ export class ItemLock extends ItemBase<'lock'> {
 		if (this.asset.definition.password != null && !playerRestrictionManager.forceAllowItemActions()) {
 			if (password == null) {
 				return null;
-			} else if (this.lockData.hidden?.side === 'server' && password !== this.lockData.hidden.password) {
+			}
+			if (this.lockData.hidden?.side === 'server' && !ItemLock._isEqualPassword(this.asset, this.lockData.hidden.password, password)) {
 				failure({
 					type: 'lockInteractionPrevented',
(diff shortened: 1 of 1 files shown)


ClaudiaMia (approved):
Tested & LGTM

pandora-common/src/assets/item.ts
```diff
@@ -1276,6 +1277,14 @@ export class ItemLock extends ItemBase<'lock'> {
 		}
 		return true;
 	}
+
+	private static _isEqualPassword(_asset: Asset<'lock'>, lhs?: string, rhs?: string): boolean {
+		if (lhs == null || rhs == null)
+			return lhs === rhs;
+
+		// all passwords are case insensitive for now
+		return lhs.localeCompare(rhs, undefined, { sensitivity: 'base' }) === 0;
```
Jomshir98, line 1286: I would really like to avoid any `locale*` function in the common - whole principle of `locale*` functions is that they behave differently based on the machine/browser's locale, which we definitely don't want for checks like this.
I see two solutions:
- Instead of `undefined` use some specific locale (the second argument)
- Use a simple way of doing `.toLowerCase()` on both strings.

Jomshir98 (changes requested):

Jomshir98 (approved):

PR #501, opened 2023-12-31
[FIX] warn user if ban/kick would fail due to the target being an admin

### LINKED ISSUE #492: Improve behavior when account is both admin and banned in a room
Currently, a room admin can be (maybe accidentally) be added to the ban list.

This has no effect in my tests, as the admin role seems to win.

I am not sure this is good behavior, as it is a bit hidden. I believe the behavior should maybe be changed so that:

- when you try to add an account as admin which is banned, open a popup to ask if it should be added as admin and removed from the ban list.
- when you try to ban an account that is admin, open a popup and ask if it should be banned and removed from the admin list.

Sekkmer:
fixes #492

(admins who are only admins due to development auto admin feature can be banned and kicked)


ClaudiaMia (changes requested):
Tested and works.

One note though: While there is a toast when you try to kick/ban via a chat command, I would have expected that same toast to also pop up when you try to click the striked-through entry n the character context menu as that normally there is always a toast when you click on something striked-through in other cases I think.

ClaudiaMia (approved):

PR #502, opened 2023-12-31
[FEATURE] mark edited message, scroll on mode change

### LINKED ISSUE #107: Highlight currently edited message
if a message is being edited, it should be visible to the user which one it is

### LINKED ISSUE #271: Roll down chat on mode change
Changing chat mode should roll down the chat to keep it on the bottom, if it was already scrolled to the bottom.

Sekkmer:
resolves #271
resolves #107

### DIFF
--- a/pandora-client-web/src/components/chatroom/chat.tsx
+++ b/pandora-client-web/src/components/chatroom/chat.tsx
@@ -79,17 +79,19 @@ function DisplayUserMessage({ message, playerId }: { message: IChatRoomMessageCh
 		}
 	}, [message.type]);
 	const [ref, onContextMenu, close] = useContextMenu();
+	const { editing } = useChatInput();
+	const editingClass = editing?.target === message.id ? 'editing' : undefined;
 
 	const style = message.type === 'me' || message.type === 'emote' ? ({ backgroundColor: message.from.labelColor + '44' }) : undefined;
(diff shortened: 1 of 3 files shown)


ClaudiaMia (commented):
Tested and works as expected.

However, I noticed two weird things:

1. The commits of this PR also include the commits from your other PR: https://github.com/Project-Pandora-Game/pandora/pull/501
2. When I checked out this branch, all stored room devices in the room inventory got automatically deployed, which is clearly a bug!
I assume this has to do with me having had checked out that other PR which introduces persistent room coordinates just before?

Sekkmer:
> Tested and works as expected.
> 
> However, I noticed two weird things:
> 
> 1. The commits of this PR also include the commits from your other PR: [[FIX] warn user if ban/kick would fail due to the target being an admin #501](https://github.com/Project-Pandora-Game/pandora/pull/501)
> 2. When I checked out this branch, all stored room devices in the room inventory got automatically deployed, which is clearly a bug!
>    I assume this has to do with me having had checked out that other PR which introduces persistent room coordinates just before?

1. I removed the commits that did not belong here

2. I currently deployed the room device change on the server, which adds a `deployed: z.boolean(true)` and the deployment is no longer nullable, so old client would think all devices are deployed, this is normal for this branch

ClaudiaMia (approved):

PR #503, opened 2023-12-31
[ADD] coloring permission of devices and minimum account password length

ClaudiaMia:
Reasoning:

~~Change 1. Give warning to users to not use a sensitive username since they think it is hidden~~ EDIT: seems with the activities around using `displayName` everywhere, this is no longer needed, so I reverted this change again
Change 2. Prevent bad actors from undoing lots of work by changing room device colors randomly without leaving a trace even in the chat
Change 3. We still should force users to select a strong enough password, despite our account security measures. Yes, this will lead to serious bad actors focusing on trying passwords with a length of 8-10 characters mainly, but overall it still improves the security situation on average, I think.

### DIFF
--- a/pandora-common/src/validation.ts
+++ b/pandora-common/src/validation.ts
@@ -178,4 +178,5 @@ export const SIMPLE_TOKEN_LENGTH = 6;
 export const SimpleTokenSchema = z.string().length(SIMPLE_TOKEN_LENGTH).regex(/^[0-9]+$/);
 export const IsSimpleToken = ZodMatcher(SimpleTokenSchema);
 
+// TODO: set minimum password length to maybe 8 characters?
 export const PasswordSha512Schema = z.string().regex(/^[a-zA-Z0-9+/]{86}==$/);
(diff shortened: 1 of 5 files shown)


pandora-common/src/validation.ts
```diff
@@ -178,4 +178,5 @@ export const SIMPLE_TOKEN_LENGTH = 6;
 export const SimpleTokenSchema = z.string().length(SIMPLE_TOKEN_LENGTH).regex(/^[0-9]+$/);
 export const IsSimpleToken = ZodMatcher(SimpleTokenSchema);
 
+// TODO: set minimum password length to maybe 8 characters?
```
Sekkmer, line 181: Password length cannot be enforced here, server does not have any information about the password length, TODO could be put in limits maybe

or better yet added as validation in `registrationForm.tsx` and `resetPaasswordForm.tsx`

ClaudiaMia: I just pushed some code that shows that I can enforce password length here together with some client code in those two files you mentioned.

It worked in my tests. Is it not correct like I did in my latest commit?

Sekkmer (approved):

PR #504, opened 2023-12-31
[CHANGE] crypto key should not be changed directly

Sekkmer:
prevent direct update of the crypto key after its first creation, as it was never an intended behavior

change DMs to refer to the upcoming `displayName` of the account instead of the username

### DIFF
--- a/pandora-server-directory/src/account/account.ts
+++ b/pandora-server-directory/src/account/account.ts
@@ -36,6 +36,11 @@ export class Account {
 		return this.data.username;
 	}
 
+	/** TODO: Allow choosing custom display name instead of always showing a username */
+	public get displayName(): string {
+		return this.data.username;
+	}
+
 	constructor(data: DatabaseAccountWithSecure) {
 		this.lastActivity = Date.now();
 		// Shallow copy to preserve received data when cleaning up secure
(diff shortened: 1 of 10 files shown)


pandora-server-directory/src/account/account.ts
```diff
@@ -89,8 +95,7 @@ export class Account {
 	public getAccountPublicInfo(): AccountPublicInfo {
 		return {
 			id: this.data.id,
-			// TODO: Allow choosing custom display name instead of always showing a username
-			displayName: this.data.username,
+			displayName: this.displayName,
```
ClaudiaMia, line 98: So I assume this means that the client no longer gets the full data package anywhere in pandora code that includes the username, but now only gets the display name from the database?

What is the default value for the display name on a new account? 
I assume it is maybe simply the account number?
Is there also a way to set this display name somewhere, like under settings > account?
Will we allow users to change the display name at any point in time and how often they want or can they only set one once?

Sekkmer: Sadly no, the client still gets the username with relationships (contracts, blocks), and that is a bit more complex to remove, since a copy of the username is already stored in the database

I just basically moved the todo to a getter property in the account that returns the username the same way as before

ClaudiaMia: Sekkmer opened a new issue: https://github.com/Project-Pandora-Game/pandora/issues/505

PR #507, opened 2024-01-02
[CHANGE] chat edit styling and small preview images as default

ClaudiaMia:


pandora-common/src/networking/directory_client.ts
```diff
@@ -60,7 +60,7 @@ export const DirectoryAccountSettingsSchema = z.object({
 	/**
 	 * Controls whether to show the attribute icons or preview images in small preview.
 	 */
-	wardrobeSmallPreview: z.enum(['icon', 'image']).default('icon'),
+	wardrobeSmallPreview: z.enum(['icon', 'image']).default('image'),
```
Jomshir98, line 63: Changing it here is not enough, update `ACCOUNT_SETTINGS_DEFAULT` lower, too.

ClaudiaMia: Oh my~ Two defaults!

Thank you, fixed it~

Jomshir98 (approved):

PR #508, opened 2024-01-03
[FIX] migrate directMessages to have the correct new displayName property

Sekkmer:
migrate to respect type change from #504
migrate account relationships to account contacts to respect UI changes in all types and database

prepare account display name, no copy of username is stored outside of account in db

the migration will be reverted before merge


pandora-server-directory/src/database/mongoDb.ts
```diff
@@ -518,6 +518,22 @@ export default class MongoDatabase implements PandoraDatabase {
 
 	private async _doMigrations(): Promise<void> {
 		// insert migration code here
+		for await (const account of this._accounts.find().stream()) {
+			if (account.directMessages == null)
+				continue;
+
+			const directMessages: typeof account.directMessages = [];
+			for (const dm of account.directMessages) {
+				const migrated = {
```
Jomshir98, line 527: Sooo... here you create `migrated`... and then?

Jomshir98 (changes requested):
Would be really nice if you tested code like this locally first.

pandora-server-directory/src/database/mongoDb.ts
Sekkmer: ...and then it disappears...

pandora-server-directory/src/account/accountDirectMessages.ts
```diff
@@ -36,6 +32,27 @@ export class AccountDirectMessages {
 		this._dms = data ?? [];
 	}
 
+	public async getDirectMessageInfo(): Promise<IDirectoryDirectMessageInfo[]> {
```
Jomshir98, line 35: This method should be async synchronized to avoid extra queries while one is already pending.

pandora-server-directory/src/account/accountDirectMessages.ts
```diff
@@ -69,6 +85,27 @@ export class AccountDirectMessages {
 				delete dm.closed;
 				break;
 		}
+		if (this._infos) {
```
Jomshir98, line 88: This can lead to a race condition with `getDirectMessageInfo`. Both should use an `object` async synchronization.

pandora-server-directory/src/database/mongoDb.ts
```diff
@@ -518,9 +519,58 @@ export default class MongoDatabase implements PandoraDatabase {
...
+		if (await this._accountContacts.countDocuments() > 0) {
+			logger.warning('Account contacts already migrated');
+			return;
+		}
+
+		const relationships = this._db.collection<DatabaseAccountRelationshipOld>('relationships');
+		const relationshipsCount = await relationships.countDocuments();
+
+		for await (const rel of relationships.find().stream()) {
+			const contact = {
```
Jomshir98, line 551: Nit: Please add an explicit type to make sure it matches exactly~ (I would have loved to see running it through a Zod schema the most, but this specific thing is only a type, not a schema right now).

Jomshir98 (changes requested):

Jomshir98 (approved):

PR #512, opened 2024-01-08
Feature/account display name

### LINKED ISSUE #505: do not share account username with other users, use display name instead
Relationships store a copy of the username, this will need to be updated on the display name change

there should be a limit on how frequently the display name can be changed, maybe once every week/month

(code should be checked if all possible paths of username sharing is removed)

Sekkmer:
resolves #505


pandora-common/src/utility.ts
```diff
@@ -200,6 +200,36 @@ export function FormatTimeInterval(time: number, mode: 'full' | 'short' = 'full'
...
+		case 'hours':
+			return time * 1000 * 60 * 60;
+		case 'days':
+			return time * 1000 * 60 * 60 * 24;
+		case 'weeks':
+			return time * 1000 * 60 * 60 * 24 * 7;
+		case 'months':
+			return time * 1000 * 60 * 60 * 24 * 30;
+		case 'years':
+			return time * 1000 * 60 * 60 * 24 * 365;
```
Jomshir98, line 229: I don't think we want to allow `months`/`years` - they are both imprecise - it would be better to define the values in code in days if such high values are wanted (same as we have reverse support for in `FormatTimeInterval`)

pandora-common/src/networking/directory_client.ts
```diff
@@ -101,6 +101,40 @@ export const ACCOUNT_SETTINGS_DEFAULT = Object.freeze<IDirectoryAccountSettings>
 	interfaceChatroomOfflineCharacterFilter: 'ghost',
 });
 
+export const DirectoryAccountSettingsLimitedSchema = z.object({
+	displayName: DisplayNameSchema,
+});
+export type IDirectoryAccountSettingsLimited = z.infer<typeof DirectoryAccountSettingsLimitedSchema>;
```
Jomshir98, line 107: The `I` prefix is inconsistent throughout the codebase, so I'm slowly trying to phase it out. I think it is better without it:
```suggestion
export type DirectoryAccountSettingsLimited = z.infer<typeof DirectoryAccountSettingsLimitedSchema>;
```

pandora-common/src/networking/directory_client.ts
```diff
@@ -101,6 +101,40 @@ export const ACCOUNT_SETTINGS_DEFAULT = Object.freeze<IDirectoryAccountSettings>
...
+
+type DirectoryAccountSettingsLimitedStoredSchemaAcc = {
+	[K in keyof IDirectoryAccountSettingsLimited]: ZodObject<{
+		value: typeof DirectoryAccountSettingsLimitedSchema.shape[K];
+		nextAllowedChange: ZodNumber;
+	}>;
+};
+export const DirectoryAccountSettingsLimitedStoredSchema = z.object(
+	KnownObject.entries(DirectoryAccountSettingsLimitedSchema.shape).reduce(
```
Jomshir98, line 116: Avoid using `reduce` where the reduced value actually never changes - using a much simpler `for` will have almost identical code.

pandora-common/src/networking/directory_client.ts
```diff
@@ -101,6 +101,40 @@ export const ACCOUNT_SETTINGS_DEFAULT = Object.freeze<IDirectoryAccountSettings>
 	interfaceChatroomOfflineCharacterFilter: 'ghost',
 });
 
+export const DirectoryAccountSettingsLimitedSchema = z.object({
+	displayName: DisplayNameSchema,
+});
+export type IDirectoryAccountSettingsLimited = z.infer<typeof DirectoryAccountSettingsLimitedSchema>;
+
+type DirectoryAccountSettingsLimitedStoredSchemaAcc = {
```
Jomshir98, line 109: Optional suggestion:
This complexity could be avoided altogether, if the "next allowed change" data was next to the values instead of interleaved with them, i.e.:
```ts
type IDirectoryAccountInfo = {
    // ...
    settingsLimited: IDirectoryAccountSettingsLimited;
    settingsLimitedCooldowns: Partial<Record<keyof IDirectoryAccountSettingsLimited, number>>;
    // ...
}
```

pandora-common/src/networking/directory_client.ts
```diff
@@ -101,6 +101,40 @@ export const ACCOUNT_SETTINGS_DEFAULT = Object.freeze<IDirectoryAccountSettings>
...
+	),
+);
+export type IDirectoryAccountSettingsLimitedStored = z.infer<typeof DirectoryAccountSettingsLimitedStoredSchema>;
+
+export const ACCOUNT_SETTINGS_LIMITED_STORED_DEFAULT = Object.freeze<IDirectoryAccountSettingsLimitedStored>({
+	displayName: { value: '', nextAllowedChange: 0 },
+});
+
+export const ACCOUNT_SETTINGS_LIMITED_LIMITS = Object.freeze<Readonly<Record<keyof IDirectoryAccountSettingsLimited, number>>>({
+	displayName: TimeSpanMs(1, 'months'),
```
Jomshir98, line 135: [discussion] I think allowing a change only once per month might be too high for this. How about once per week?
As I said before I honestly think the limiting is not needed at all. The action of changing your name has no performance cost for us and for users it will be similarly confusing if user changes the name once as as if they change it multiple times - they always see only the last value either way.
Only case this affects is when person A changes their name, person B sees it and _gets used to the new one_, and then A changes the name again. This will however be problem no matter what time period we choose. On other hand long period might as well be almost the same as not allowing the name change at all.

pandora-common/src/networking/directory_client.ts
```diff
@@ -101,6 +101,40 @@ export const ACCOUNT_SETTINGS_DEFAULT = Object.freeze<IDirectoryAccountSettings>
...
+			return acc;
+		},
+		// eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+		{} as DirectoryAccountSettingsLimitedStoredSchemaAcc,
+	),
+);
+export type IDirectoryAccountSettingsLimitedStored = z.infer<typeof DirectoryAccountSettingsLimitedStoredSchema>;
+
+export const ACCOUNT_SETTINGS_LIMITED_STORED_DEFAULT = Object.freeze<IDirectoryAccountSettingsLimitedStored>({
+	displayName: { value: '', nextAllowedChange: 0 },
```
Jomshir98, line 131: This default value isn't actually valid - the `DisplayNameSchema` specifies minimum length of 3.

pandora-server-directory/src/database/mongoDb.ts
```diff
@@ -255,15 +255,14 @@ export default class MongoDatabase implements PandoraDatabase {
...
-			.project<Pick<DatabaseAccount, 'id' | 'username'>>({ id: 1, username: 1 })
+			.project<Pick<DatabaseAccount, 'id' | 'settingsLimited'>>({ id: 1, settingsLimited: 1 })
 			.toArray();
 
 		for (const acc of accounts) {
-			result[acc.id] = acc.username;
+			result[acc.id] = acc.settingsLimited.displayName.value;
 		}
 		return result;
 	}
```
Jomshir98, line 268: This needs to default to username if the display name is empty, as we shouldn't allow empty names anywhere (because if the `settingsLimited` gets invalid somehow, they get reset into default value which _is_ empty in the current implementation (which is a bug as after reset it will immediately fail the validation, because that is not a valid value)).

pandora-server-directory/src/database/mongoDb.ts
```diff
@@ -519,6 +518,12 @@ export default class MongoDatabase implements PandoraDatabase {
 
 	private async _doMigrations(): Promise<void> {
 		// insert migration code here
+		for await (const account of this._accounts.find()) {
+			const settingsLimited = _.cloneDeep(ACCOUNT_SETTINGS_LIMITED_STORED_DEFAULT);
+			settingsLimited.displayName.value = account.username;
```
Jomshir98, line 523: This shouldn't happen if the display name is already set.

pandora-common/src/networking/directory_client.ts
Jomshir98: Actually thinking about it more...
Why is this a completely separate settings framework in the first place when the following would have been enough to add limited settings _if we actually want them in the first place_?
```ts
// Add the `displayName` to the existing settings

const ACCOUNT_SETTINGS_LIMITS: Partial<Record<keyof IDirectoryAccountSettings, number>> = {
    displayName: TimeSpanMs(1, 'months'),
};

type IDirectoryAccountInfo = {
    // ...
    settingsCooldowns: Partial<Record<keyof IDirectoryAccountSettings, number>>;
    // ...
}
```

pandora-server-directory/src/account/accountContacts.ts
```diff
@@ -293,43 +296,60 @@ export class AccountContacts {
...
+		this._updatingStatus = true;
+		try {
+			await this.load();
+		} finally {
+			this._updatingStatus = true;
+			this._updateStatus();
+		}
+	}
+
+	public _updateStatus(): void {
```
Jomshir98, line 352: Is this meant to be `public`?

Jomshir98 (changes requested):
I like the contacts changes removing the cached name and instead querying it.

pandora-common/src/networking/directory_client.ts
Sekkmer: I have no idea why this would be beneficial, considering that we would need the for to be in global scope or an immediately invoked lambda, non of which seems better for me

pandora-server-directory/src/account/account.ts
```diff
@@ -115,11 +115,34 @@ export class Account {
 			settings.visibleRoles = uniq(settings.visibleRoles.filter((role) => this.roles.isAuthorized(role)));
 		}
 
+		const db = GetDatabase();
+		const update: Parameters<(typeof db)['updateAccountData']>['1'] = {};
+
+		const now = Date.now();
+		for (const [key, value] of KnownObject.entries(this.data.settingsCooldowns)) {
```
Jomshir98, line 122: This should iterate over `ACCOUNT_SETTINGS_LIMITED_LIMITS` keys instead, because the `data.settingsCooldowns` is partial - it could skip keys if they were never set.
The main bit is that `settingsCooldowns` write needs to happen, which it currently doesn't for a new account.

Maybe it would be simpler iterating over the new `settings` object and checking if the key is a limited setting?

pandora-client-web/src/components/settings/accountSettings.tsx
```diff
@@ -198,3 +200,43 @@ function LabelColor({ account }: { account: IDirectoryAccountInfo; }): ReactElem
...
+	const [name, setName] = useState(current);
+
+	const onSetDisplayName = useEvent(() => {
+		if (name === current)
+			return;
+		if (nextAllowedChange && nextAllowedChange > Date.now()) {
+			toast(`You can change your display name again in ${FormatTimeInterval(nextAllowedChange - Date.now())}`, TOAST_OPTIONS_ERROR);
+			return;
+		}
+		const displayName = account.username === current ? null : current;
```
Jomshir98, line 217: This should use the new name, not the old one...
```suggestion
		const displayName = name === current ? null : name;
```

pandora-common/src/networking/directory_client.ts
```diff
@@ -101,6 +103,13 @@ export const ACCOUNT_SETTINGS_DEFAULT = Object.freeze<IDirectoryAccountSettings>
 	interfaceChatroomOfflineCharacterFilter: 'ghost',
 });
 
+export const ACCOUNT_SETTINGS_LIMITED_LIMITS = Object.freeze({
+	displayName: TimeSpanMs(1, 'weeks'),
+} as const satisfies Partial<Record<keyof IDirectoryAccountSettings, number>>);
+
+export type DirectoryAccountSettingsLimitedKeys = keyof typeof ACCOUNT_SETTINGS_LIMITED_LIMITS;
+export type DirectoryAccountSettingsCooldowns = Partial<Record<DirectoryAccountSettingsLimitedKeys, number>>;
```
Jomshir98, line 111: These should be schemas to be validate-able:
```suggestion
export const DirectoryAccountSettingsLimitedKeysSchema = z.enum(ParseArrayNotEmpty(KnownObject.keys(ACCOUNT_SETTINGS_LIMITED_LIMITS)));
export type DirectoryAccountSettingsLimitedKeys = z.infer<typeof DirectoryAccountSettingsLimitedKeysSchema>;

export const DirectoryAccountSettingsCooldownsSchema = z.record(DirectoryAccountSettingsLimitedKeysSchema, z.number().optional());
export type DirectoryAccountSettingsCooldowns = z.infer<typeof DirectoryAccountSettingsCooldownsSchema>;
```

pandora-server-directory/src/database/databaseStructure.ts
```diff
@@ -66,6 +67,7 @@ export const DatabaseAccountSchema = z.object({
 	profileDescription: z.string().default('').transform(ZodTruncate(LIMIT_ACCOUNT_PROFILE_LENGTH)),
 	characters: ZodCast<ICharacterSelfInfoDb>().array(),
 	settings: DirectoryAccountSettingsSchema.catch(() => cloneDeep(ACCOUNT_SETTINGS_DEFAULT)),
+	settingsCooldowns: ZodCast<DirectoryAccountSettingsCooldowns>().catch(() => ({})),
```
Jomshir98, line 70: This needs to be a schema for the migration to actually happen - `ZodCast` simply ignores the value and passes it along as-is - same as using `as` to cast the type - so the `catch` will never trigger.
```suggestion
	settingsCooldowns: DirectoryAccountSettingsCooldownsSchema.catch(() => ({})),
```

Jomshir98 (changes requested):

Jomshir98 (approved):

PR #517, opened 2024-01-12
[ADD] asset preference and item coloring permissions

Sekkmer:


pandora-common/src/character/restrictionsManager.ts
```diff
@@ -296,10 +297,17 @@ export class CharacterRestrictionsManager {
...
-				case 'favorite':
-				case 'normal':
 				case 'maybe':
+					context.addInteraction(target.character, 'assetPrefMaybe');
+				// Fallthrough
+				case 'normal':
+					context.addInteraction(target.character, 'assetPrefNormal');
+				// Fallthrough
+				case 'favorite':
+					context.addInteraction(target.character, 'assetPrefFavorite');
```
Jomshir98, line 307: Should they really be fallthrough? I'm not fully sure~

pandora-common/src/gameLogic/interactions/_interactionConfig.ts
```diff
@@ -36,12 +42,30 @@ export const INTERACTION_CONFIG = {
...
+		defaultPermissions: {
+			allowOthers: true,
+		},
+	},
+	assetPrefMaybe: {
+		visibleName: 'Add and remove assets with maybe preference',
+		defaultPermissions: {
+			allowOthers: true,
+		},
+	},
```
Jomshir98, line 62: I'm not sure about these being part of interactions config... I would rather move them to be their own category. If you wouldn't mind I can implement that as I have a good idea how to do that. (I intended for the solution to be used that way, but there is so far no example of how I actually imagined it being done)

pandora-common/src/assets/appearanceActions.ts
```diff
@@ -366,22 +328,14 @@ export function DoAppearanceAction(
...
-			if (!r.allowed) {
-				processingContext.addProblem({
-					result: 'restrictionError',
-					restriction: r.restriction,
-				});
+			processingContext.checkCanUseItem(target, action.item, ItemInteractionType.STYLING);
+
+			if (target.type === 'character') {
+				processingContext.addInteraction(target.character, 'changeItemColor');
 			}
```
Jomshir98, line 339: Thought: Maybe worth moving this interaction check into `canUseItem` when the interaction type is `STYLING`? That type is only used for coloring right now (and in theory for modules that mark themselves as styling, which makes sense for some modules like light/dark variant).

Jomshir98 (changes requested):

pandora-common/src/gameLogic/assetPreferences/assetPreferencesSubsystem.ts
```diff
@@ -0,0 +1,48 @@
...
+} as const satisfies Immutable<Record<AssetPreferenceType, AssetPreferencePermissionConfig | null>>;
+
+export interface AssetPreferencePermissionConfig {
+	visibleName: string;
+	defaultPermissions: PermissionConfigDefault;
+}
+
+export abstract class AssetPreferencesSubsystem extends TypedEventEmitter<AssetPreferencesSubsystemEvents> implements IPermissionProvider {
+	public abstract get currentPreferences(): Immutable<AssetPreferencesPublic>;
+	public abstract getPreferencePermission(preference: AssetPreferenceType): GameLogicPermission | null;
```
Sekkmer, line 42: This feels like a complete duplication of the other Subsystem, but for some reason, this function can return null, the other throws an exception.

What exactly do we gain from this much extra duplicated code, if we just want to split it for the UI, a simple group in the config would have been able to accomplish the same thing with much less code

Sekkmer: well I tried this but over all it did not make anything better, the duplication should be fine for now, I guess

Jomshir98 (approved):

PR #519, opened 2024-01-12
[ADD] eslint to warn about @deprecated

EveFerox:


Jomshir98 (changes requested):
`jest.config.ts` definitely shouldn't be added to the main tsconfig(s). This will cause it to be included in the compilation of production code, which we definitely don't want.
If the deprecation plugin needs it, then either the file should be added to the `test/tsconfig.json`, or (my preference) the plugin should be moved to the `overrides` area of `eslintrc.json`. The overrides are are rules that only apply to `.ts(x)` files, so it could be edited to be something like:
```json
"extends": [
	"plugin:@typescript-eslint/recommended-type-checked",
	"plugin:@typescript-eslint/stylistic-type-checked",
	"deprecation"
],

File: pandora-common/package.json [hunk 1]
```diff
@@ -49,6 +49,7 @@
 		"@typescript-eslint/eslint-plugin": "6.18.1",
 		"@typescript-eslint/parser": "6.18.1",
 		"eslint": "8.56.0",
+		"eslint-plugin-deprecation": "^2.0.0",
```
Jomshir98, line 52: Please use exact version for dependency:
```suggestion
		"eslint-plugin-deprecation": "2.0.0",
```
And then run `pnpm i` and commit both this change and updated lockfile.

Jomshir98:
The `execCommand` deprecations can be safely ignored by adding the following comment above them:
```ts
// eslint-disable-next-line deprecation/deprecation
```
They are intentionally a fallback for browsers that don't allow the preferred way if copying into clipboard.

EveFerox:
> `jest.config.ts` definitely shouldn't be added to the main tsconfig(s). This will cause it to be included in the compilation of production code, which we definitely don't want. If the deprecation plugin needs it, then either the file should be added to the `test/tsconfig.json`, or (my preference) the plugin should be moved to the `overrides` area of `eslintrc.json`. The overrides are are rules that only apply to `.ts(x)` files, so it could be edited to be something like:
> 
> ```json
> "extends": [
> 	"plugin:@typescript-eslint/recommended-type-checked",
> 	"plugin:@typescript-eslint/stylistic-type-checked",
> 	"deprecation"
> ],
> ```

tried a few .eslintrt.json variations but they all fail at `pnpm run lint` with:
```
ESLint couldn't find the config "deprecation" to extend from. Please check that the name of the config is correct.
```

Jomshir98 (approved):
Looks good to me now! Feel free to merge if you are happy with it, too.

PR #515, opened 2024-01-10
[ADD] debug option to rate limit observables

Sekkmer:

PR #521, opened 2024-01-13
[FIX] Properly override `Observable`'s `value` for `RateLimitedObservable`

Jomshir98:
Seems final version of #515 was not tested, because it completely breaks development and userdebug builds.
This is because in ECMAScript spec overriding getter/setter sets a new property on the current prototype. This property, however, overrides both set and get actions from the parent prototype. This means that to get the expected results both need to be overridden at the same time.


Sekkmer (approved):
Ehh, why must JavaScript always need to come up with ways to be a mess, and TypeScript to lie about it...

PR #523, opened 2024-01-13
[CHANGE] use integer-based accurate conversion for RGB/HSV

Sekkmer:
code is based on:
https://sci-hub.se/10.1016/j.compeleceng.2015.08.005


Sekkmer:
some HSV/HSL conversion got me confused, should be fixed now

PR #524, opened 2024-01-14
[FIX] use integer-based accurate conversion for RGB/HSV

Sekkmer:
Fixes the mouse position calculation inside the square so the circle is no longer slightly off

code is based on:
https://sci-hub.se/10.1016/j.compeleceng.2015.08.005

Note:
The linear gradient expands outside the range of the possible slider position, which causes some inaccuracy. 
This is mostly noticeable for the HUE slider.
This was an issue before the change; I will address it later in PR.


Jomshir98 (commented):
Looks good and mainly feels a lot better!

PR #527, opened 2024-01-15
[FIX] stable random actions

### LINKED ISSUE #511: use stable random generation for appearance actions, with seed as a parameter of the action
All appearance actions should always have the same results if the inputs are the same

`Math.random()` should be replaced to use our `PseudoRandom` class

currently this causes a rapid preview change when the user hovers over
Wardrobe -> Randomization -> any of the 2 buttons with preview enabled

Sekkmer:
fixes #511

fix issues with hover previews by disabling observable rate limits by default


pandora-client-web/src/browserStorage.ts
```diff
@@ -38,23 +38,14 @@ export class BrowserStorage<T> extends Observable<T> {
...
-				}
+			const validated = this.validate.safeParse(parsedValue);
+			if (validated.success) {
+				this.value = validated.data;
 			}
 		}
 	}
 
-	public static create<T>(name: string, defaultValue: T, validate?: ZodType<T, ZodTypeDef, unknown> | ((value: unknown) => boolean)): BrowserStorage<T> {
+	public static create<T>(name: string, defaultValue: T, validate: ZodType<T, ZodTypeDef, unknown>): BrowserStorage<T> {
```
Jomshir98, line 48: In this function the `validate ?? (() => true)` can be simplified to `validate`.

Sekkmer: I wish TS would scream from an invalid type on both side of `??` regardless if the first part is always non null

pandora-client-web/src/ui/components/chat/chatInput.tsx
```diff
@@ -46,11 +46,12 @@ export type IChatInputHandler = {
 
 const chatInputContext = createContext<IChatInputHandler | null>(null);
 
-type ChatInputSave = {
-	input: string;
-	spaceId: SpaceId | null;
-};
-const InputRestore = BrowserStorage.createSession<ChatInputSave>('saveChatInput', { input: '', spaceId: null });
+const ChatInputSchema = z.object({
```
Jomshir98, line 49:
```suggestion
const ChatInputSaveSchema = z.object({
```

PR #532, opened 2024-01-19
[ADD] icons for all individual permissions

ClaudiaMia:


Sekkmer (approved):

ClaudiaMia:
There are two options to go for here:

1. Writing yes/ask/no for the permission configuration state, which is nicely descriptive for new players
2. Using icons, which I prefer as I feel it gives more long-term value as it gives a better comparison and overview. What they mean can be read anyway, after clicking on edit.

I reverted back to option (2). Thoughts?

preview

Sekkmer:
I prefer version 2 with icons

Sekkmer (approved):

PR #533, opened 2024-01-19
[FEATURE] item limits

### LINKED ISSUE #453: Item limit handling
Potential beta 1 candidate, as it is good to add limits as early as possible, since later on, adding limits may feel bad.

There is currently a limit of how many items can be in a room inventory, including items nested inside various layers of containers.

There should also be a limit added of how many items a character can wear on their body.
I propose two limits:
One limit for wearing items on the highest layer.
A second limit for how many nested items could be carried in containers on the character.

The reason for splitting it into two limits is that we want to likely limit how much of their budget someone can spend on creating a visible  and rendered outfit.
1. for performance reasons
2. for the reason of managing the maximum slowdown if we would add a remove/add queue to actions with a tiny slow down (Note: Body item changes should likely be instant and not be queued).

Both limits should not be set too big at the beginning so we can test rendering performance in large rooms, and so that we can see the impact of being able to carry many items in a bag. As we later on plan to have a room admin configuration to limit item spawning in the room, being able to carry hundreds of items into the room in bags would render that feature potentially useless, unless you are not allowed to enter with storage items, too.

Also make those limits transparent to the users in the UI, similar to how it was done for outfits.

Sekkmer:
resolves #453


pandora-client-web/src/components/wardrobe/wardrobeItems.tsx
```diff
@@ -101,10 +104,14 @@ export function WardrobeItemManipulation({ className }: { className?: string; })
 	const currentFocus = useObservable(focus);
 	const { preFilter, containerContentsFilter, assetFilterAttributes } = useWardrobeItems(currentFocus);
 
-	const title: string = target.type === 'character' ? 'Currently worn items' : 'Room inventory';
+	const appearance = useWardrobeTargetItems(target);
+	const itemCount = useMemo(() => AppearanceItemsCalculateTotalCount(appearance), [appearance]);
+	const itemLimit: number = target.type === 'character' ? ITEM_LIMIT_CHARACTER_WORN : ITEM_LIMIT_ROOM_INVENTORY;
+	const titlePrefix: string = target.type === 'character' ? 'Currently worn items' : 'Room inventory';
+
+	const title = `${ titlePrefix }, used: ${ itemCount } / ${ itemLimit } (${100 * itemCount / itemLimit}%)`;
```
Jomshir98, line 112: Same thing about the ceil:
```suggestion
	const title = `${ titlePrefix }, used: ${ itemCount } / ${ itemLimit } (${  Math.ceil(100 * itemCount / itemLimit) }%)`;
```

pandora-common/src/assets/validation.ts
```diff
@@ -0,0 +1,67 @@
+import type { ItemId } from './appearanceTypes';
```
Jomshir98, line 1: This file seems to be missing in the `/src/assets/index.ts` file. Is that intentional?

pandora-common/src/assets/appearanceValidation.ts
```diff
@@ -60,6 +60,17 @@ export type AppearanceValidationResult = {
...
+export function AppearanceItemsCalculateTotalCount(items: AppearanceItems): number {
+	let itemCount = 0;
+	for (const item of items) {
+		itemCount++;
+		for (const module of item.getModules().values()) {
+			itemCount += AppearanceItemsCalculateTotalCount(module.getContents());
+		}
+	}
+	return itemCount;
+}
```
Jomshir98, line 72: Nit: I feel this function makes more sense to be put in `./validation.ts` file. Also avoids cyclic import.

Jomshir98 (changes requested):

Jomshir98 (approved):
LGTM!

PR #534, opened 2024-01-19
[ADD] mitigation for timing based attack for sensitive functions

Sekkmer:


pandora-server-directory/src/networking/manager_client.ts
```diff
@@ -828,3 +834,35 @@ async function TestCaptcha(token?: string): Promise<boolean> {
...
+		try {
+			result = [true, await fn(...args)];
+		} catch (e) {
+			result = [false, e];
+		}
+		const after = Date.now();
+		const elapsed = after - before;
+		const remaining = delay - elapsed;
+		if (remaining > 0) {
+			await new Promise((resolve) => setTimeout(resolve, remaining));
```
Jomshir98, line 861: Nitpick: We have an existing `Sleep` utility function.

Jomshir98 (approved):
Looks good and works! We might want some progress indicator on the login and registration pages as they now visibly lag, but that is separate issue.

PR #537, opened 2024-01-20
[FEATURE] permission prompt and character based overrides

Sekkmer:


pandora-client-web/src/components/settings/permissionsSettings.tsx
```diff
@@ -76,19 +76,62 @@ function GetIcon(icon: string): string {
...
+				return {
+					src: forbid,
+					alt: 'General permission configuration preview',
+					description: 'No one is allowed to do this, but exceptions can be set individually.',
+				};
+			case 'prompt':
+				return {
+					src: prompt,
+					alt: 'General permission configuration preview',
+					description: 'Everyone is allowed to do this, but exceptions can be set individually.',
```
ClaudiaMia, line 118: This still has the copied description. Maybe it could be something like:

"Trying to use this permission opens a popup that lets the targeted user decide if they want to give or deny the requester this permission. Exceptions can be set individually."

PR #540, opened 2024-01-23
[ADD] definition based override for asset preference

Sekkmer:
only allow "maybe" as an override

### DIFF
--- a/pandora-common/src/assets/definitions.ts
+++ b/pandora-common/src/assets/definitions.ts
@@ -11,6 +11,7 @@ import type { AssetManager } from './assetManager';
 import _ from 'lodash';
 import { BONE_MAX, BONE_MIN } from './appearance';
 import type { RoomDeviceProperties } from './roomDeviceProperties';
+import type { AssetPreferenceType } from '../character';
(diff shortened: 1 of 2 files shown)


pandora-common/src/assets/definitions.ts
```diff
@@ -110,6 +111,12 @@ export interface AssetBaseDefinition<Type extends AssetType, A extends AssetDefi
 	 *  - `undefined` temporarily until all previews are added
 	 */
 	preview?: string | null;
+
+	/**
+	 * Override the default asset preference
+	 *  - only 'maybe' is allowed since 'favorite', 'normal', 'prevent' and 'doNotRender' does not make sense
```
ClaudiaMia, line 117: Why does 'prevent' make no sense?

I feel we would want to set the default of password locks to "prevent", not to 'maybe'.

Sekkmer: wouldn't maybe be enough, I mean you would get prompted, but sure I allow prevent as well

Sekkmer: added prevent as an option too

PR #551, opened 2024-01-27
[CHANGE] chat style changes and added configurable font size setting

ClaudiaMia:


pandora-client-web/src/ui/components/chat/chat.tsx
```diff
@@ -87,7 +95,17 @@ function DisplayUserMessage({ message, playerId }: { message: IChatMessageChat &
 	const self = message.from.id === playerId;
 	return (
 		<>
-			<div className={ classNames('message', message.type, isPrivate && 'private', editingClass) } style={ style } onContextMenu={ self ? onContextMenu : undefined }>
+			<div
+				className={ classNames(
+					'message',
+					`fontSize-${interfaceChatroomChatFontSize}`,
```
Sekkmer, line 101: I would like to suggest a different approach that is more performant:
in CSS use a variable for the font size
have a hook that updates the style property and modifies the CSS variable based on the settings
with this React will only need to update 1 element instead of the possible thousands of chat messages

Sekkmer (changes requested):
DMs should also be effected by this as we have an almost identical UI there

Sekkmer (approved):

PR #553, opened 2024-01-28
[CHANGE] Updates to wiki and contribution guideline

ClaudiaMia:


ClaudiaMia:
One potential bug with the <Link to /> I added in the Permissions-tab, which leads to an article in the wiki:
When I user my browser back button afterwards, to go back to the Permissions-tab, it tells me my character is no longer logged in, likely leading to the user thinking they got disconnected, especially as only "logout" is available when you use the top-right button. The "back"-tab fixes this, but I still think this is not the intended user experience.

pandora-client-web/src/components/wiki/pages/contact.tsx
```diff
@@ -25,14 +25,20 @@ export function WikiContact(): ReactElement {
...
 				The game is open source - you can find all related sources on our <ExternalLink href='https://github.com/Project-Pandora-Game'>GitHub</ExternalLink><br />
-				If you want to contribute something (either a feature or an asset), please contact us first on our Discord,
-				so we can synchronize the efforts to make that happen.
+				If you want to contribute something (either a feature or an asset), please check
+				the <ExternalLink href='https://github.com/Project-Pandora-Game/pandora/blob/master/CONTRIBUTING.md'>contributing guideline</ExternalLink> and then
+				contact us on our Discord, so we can synchronize the efforts to make that happen.
+			</p>
+			<p>
+				For asset creators specifically, there is
+				another <ExternalLink href='https://github.com/Project-Pandora-Game/pandora-assets/blob/master/CONTRIBUTING.md'>contributing guideline</ExternalLink> for the asset repository and
```
Sekkmer, line 40: I would leave some of this commented out as assets repo is not visible publicly as of now

ClaudiaMia: Okay, pushed a change, but didn't we want to make them public before going into beta? Or do I remember that wrong?

Sekkmer: oh right good question

Sekkmer (approved):

PR #564, opened 2024-02-01
[ADD] Graphics layers for soles

Jomshir98:
Adds graphics layers for sole and a few small related QoL improvements and fixes.


pandora-client-web/src/editor/graphics/layer/setupLayer.tsx
```diff
@@ -26,9 +26,33 @@ export function SetupLayer({
...
+			}
+		}
+		return scalingBaseimage;
+	}, [evaluator, scaling, scalingBaseimage]);
+
+	const [, imageUv] = useMemo((): [string, Immutable<Record<BoneName, number>>] => {
+		const resultSetting = setting.overrides.find((img) => EvaluateCondition(img.condition, (c) => evaluator.evalCondition(c, item))) ?? setting;
+
+		return [resultSetting.image, resultSetting.uvPose ?? {}];
+	}, [evaluator, item, setting]);
```
Sekkmer, line 51: is this the same in all 3 cases, could this be a lifted into a hook?

pandora-common/src/assets/graphics/graphics.ts
```diff
@@ -207,6 +210,7 @@ export enum LayerSide {
 
 export const LayerImageSettingSchema = z.object({
 	image: z.string(),
+	uvPose: z.record(BoneNameSchema, z.number()).optional(),
```
Sekkmer, line 213: it is not quite clear for me what this will be used for, I did not see it used for soles either

maybe we should provide a small description for it

Sekkmer: if you add it later when we have editor that also fine

Jomshir98: This is an experiment that generalizes the `scaling` option, but it is unfinished and doesn't produce the best results so far - mostly on the asset side of things. I will add comments explaining that.

PR #565, opened 2024-02-01
[FEATURE] space allow list

Sekkmer:


pandora-server-directory/src/spaces/space.ts
```diff
@@ -462,7 +506,7 @@ export class Space {
 
 	/** Returns if this space is visible to the specific account when searching in space search */
 	public checkVisibleTo(account: Account): boolean {
-		return this.isValid && (this.isAdmin(account) || this.isPublic);
+		return this.isValid && (this.isPublic || this.isAdmin(account) || this.config.allow.includes(account.id));
```
Jomshir98, line 509: Might be nice to add `isAllowed(account: Account)`, similar to the existing `isAdmin`/`isBanned` helpers. It would also be implied by `isAdmin`.

pandora-server-directory/src/spaces/space.ts
```diff
@@ -360,8 +391,14 @@ export class Space {
 			return 'ok';
 
 		// If the space is full, you cannot enter it (some checks ignore space being full)
-		if (this.characterCount >= this.config.maxUsers && !ignore.characterLimit)
-			return 'errFull';
+		if (!ignore.characterLimit) {
+			let maxUsers = this.config.maxUsers;
+			if (this.isAdmin(character.baseInfo.account)) {
+				maxUsers = Math.ceil(maxUsers * LIMIT_SPACE_MAX_CHARACTER_ADMIN_OVERSHOT_MULTIPLIER);
+			}
```
Jomshir98, line 398: I'm not sure if we want this overshoot to be for admins, or only for owners. While it is not configurable only owners makes more sense to me.
Would you also please create an issue that we want to make this configurable?

Jomshir98 (changes requested):

pandora-common/src/inputLimits.ts
```diff
@@ -20,6 +20,8 @@ export const LIMIT_MAIL_LENGTH = 256;
 
 /** The maximum amount of characters inside a space */
 export const LIMIT_SPACE_MAX_CHARACTER_NUMBER = 100;
+/** The overshot for admins inside the space */
+export const LIMIT_SPACE_MAX_CHARACTER_ADMIN_OVERSHOT_MULTIPLIER = 1.1;
```
Jomshir98, line 24: Also been thinking about this - multiplier feels a bit strange for the end user. Slightly in-transparent, harder to predict and weird that size of room would affect how many admins can be in extra, especially as the increase points are relatively arbitrary from the user's point of view.

My thoughts are toward making a static, additive overshoot increase instead.

pandora-server-directory/src/spaces/space.ts
ClaudiaMia: I think that argument makes indeed sense.
With a static overshoot limit, there is a  risk an owner could not be allowed in, if there are many admins.
Making things more configurable later makes definitely sense.

pandora-common/src/inputLimits.ts
ClaudiaMia: Well, I guess if we make the change in the previous comment that it only affects owners for now, we can make it a static 5 for now, but in the future we would want to make this static limit to be user configurable, although we likely would want to still have a max defined, as otherwise you could make the room have 100 slots, plus any number of overshoot for instance for VIPs.

pandora-server-directory/src/spaces/space.ts
```diff
@@ -371,6 +408,13 @@ export class Space {
 		if (this.isBanned(character.baseInfo.account))
 			return 'noAccess';
 
+		// If you are on the allow list, you can enter the space
+		if (this.config.allow.includes(character.baseInfo.account.id))
+			return 'ok';
+
+		if (!this.isInUse())
+			return 'spaceNotInUse';
```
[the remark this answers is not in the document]
Sekkmer: no this is not specific to invites, though it is invites that give UI access to this option

currently you can enter a space that has no loaded character inside even if you are not an admin, this aims to universally prevent it, before invites you could do it by manually crafting the join with the inactive space id

pandora-server-directory/src/spaces/space.ts
Sekkmer: I made it a fixed 5 for owners

Sekkmer: also, would it make sense to limit the overshot to owners, since admins can do anything an owner could do
and to me this was to prevent a room from being filled with normal characters
like what if you fill a room with all inactive or characters who are unable to leave
both cases admin can fix the room just as much as an owner could
and considering we do not have shared ownership implemented it would make sense to allow overshot for admins and owners just the same

Jomshir98: Owners can do a bit more than admins:
- They can't be de-owned, de-admined or banned by others
- They can add more owners
So "hostile" admin (even singular) could abuse this to fill room they don't own with dummy characters until owner can't join and just leave it there. At that point the only thing owner could do would be to abandon the room from the search screen, which would mean loosing anything inside.

pandora-server-directory/src/spaces/space.ts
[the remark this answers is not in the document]
Jomshir98: Okay, I walked a bit more through the logic changes this change brings. There are a few cases where you can get to this line and if it wasn't there still be allowed to join (possibly):
- If the room is password protected and you know the password. This is removed by this change. I would argue that we should allow people that know the password to enter even if the room is not loaded, as for me password is an alternative mechanism to specifying allowlist - instead of telling Pandora who can join, you define who can join by giving them the password. So I think password and allow should be treated similarly.
- If the room is public, but has no admin inside. I honestly think this might be an existing bug and the check should use `this.isPublic` instead of `config.public`. `hasAdminInside` Also should be updated to use `c.isOnline()` instead of `c.assignedClient != null` to be more consistent.
- If the person has a valid invite, they should be able to join with it. (this change prevents that iff there is noone online in the room) On the other hand I think it would be better to have invite check that the person who created the invite is currently in the room for it to still be valid. To me that feels like a more reliable behavior than "anyone is online".

pandora-server-directory/src/spaces/space.ts
```diff
@@ -349,7 +380,7 @@ export class Space {
 		// Ignore those - the requests will fail and once the space is not requestest for a bit, it will be unloaded from the directory too, actually vanishing
 	}
 
-	public checkAllowEnter(character: Character, data: { password?: string; invite?: SpaceInviteId; }, ignore: { characterLimit?: boolean; password?: boolean; } = {}): 'ok' | 'errFull' | 'noAccess' | 'invalidPassword' | 'invalidInvite' {
+	public checkAllowEnter(character: Character, data: { password?: string; invite?: SpaceInviteId; }, ignore: { characterLimit?: boolean; password?: boolean; } = {}): 'ok' | 'errFull' | 'noAccess' | 'invalidPassword' | 'invalidInvite' | 'spaceNotInUse' {
```
Jomshir98, line 383: Whole walking through this I noticed that an option to ignore password was added and that this option is used when querying _extended_ info about an existing room.
This is against the intention the code was written with - the _extended_ info should only be available if and only if the character can join a room - it is away to "peek" into the room without having to join (e.g. to see who is currently inside). I think people that don't have a password shouldn't be allowed to peek inside a password-protected room.

pandora-server-directory/src/spaces/space.ts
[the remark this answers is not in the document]
Jomshir98: Claudia changed my mind about the second point. Public room without anyone inside should be still freely joinable so long you know the ID. It should only be unlisted on the public search.
Either way this change would prevent this.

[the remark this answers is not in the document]
ClaudiaMia: I agree with the first one. But as the space would not be listed, you would need to know the ID and use the general link or an invite link if you have one to get to the interface where you can enter the password to join.

About the 3rd point: I also think that knowing the ID of a space and using it in a general join URL can bring up the join interface without showing the extended information and if the space is public but unlisted, you should not be able to join, does not matter if no one is online, or someone is online but no admin inside. This is to protect a public space without an admin from people who may spam the chat, or disturb in other ways, e.g. by moving their character around wildly.

If you have a special invite that allows you specifically or anyone who has the invite to join the space, then I think it would be confusing though, if you can only use it to join if the one who made the invite is currently inside the space (online or not). I think if you have a special valid invite link (not the one anyone can make by copying the space ID into an URL) you should always be able to join, even if the room unlisted (empty or not), unless of course the room is password protected and your invite link does not allow to bypass that.

One thought I had in addition was that it could be nice if any person inside a public room that is not password protected could still invite someone they know through DMs to an unlisted room they are playing inside in currently. But this would not be a priority.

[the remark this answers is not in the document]
Jomshir98: The last two parts can be summarized as: You can join if you have a valid invite link.

Next would be a discussion about what is a "valid" invite link. But I would leave that for another PR, as this one focuses on adding allowlist.

Personally I would say that invite can specify if it is valid even without the one who created it is inside the space. Only admins+ would be able to create invite that is valid without them in the space, but anyone could create an invite that is valid only while they are still inside the space.
To me that makes sense as you want to be able to invite friend somewhere where you already are (of course would only work if you can join the room yourself normally, e.g. the room must be public (even if unlisted) - if it has strict whitelist-only entry, then you can't invite someone who isn't whitelisted unless you are an admin and so can add them in the first place), but we want to avoid person being able to create invite that can be used even long after they leave, which would be unpleasant surprise for those left inside the room who assume it is a private room.
Doing this might however need adjustments to how invites are limited as having constant limit of n invites per space would make it easy to DoS being able to create invites in this case.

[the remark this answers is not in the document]
ClaudiaMia: Yes, I fully agree with this last comment!

And the generic https://project-pandora.com/space/join/safgwsuzagasjasdfsafsa is from my point of view not a valid invite link to join unless the room is listed fully public with no restrictions and anyone can join anyway.

pandora-server-directory/src/spaces/space.ts
Sekkmer: ahh I see, honestly I thought this was a bug, and assumed if a room is public you should be able to see the extended info

pandora-server-directory/src/spaces/space.ts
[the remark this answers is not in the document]
Sekkmer: for this PR I limit changes to this

- public rooms can only be joined if they are publicly listed
- you can always join with password
- unique invite will be able to show extended info all the time, but password protected room need bypassPassword flag in the invite

for the first point, I am unsure if shard type stable should be checked or not, I feel natural about it, for now I inclined to not check this

pandora-server-directory/src/spaces/space.ts
Sekkmer: I changed this to `passwordOnInvite` this allows invites to work if they do not have `bypassPassword`

maybe invites could be changed to always bypass password

pandora-server-directory/src/spaces/space.ts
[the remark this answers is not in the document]
Sekkmer: also not part of this PR but UI for joining unlisted password protected rooms is missing, so the only way to do so is manual for now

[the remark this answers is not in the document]
ClaudiaMia: > also not part of this PR but UI for joining unlisted password protected rooms is missing, so the only way to do so is manual for now

I do not think there will be any UI for joining unlisted password protected spaces.
You either know the link to it or not and that is fine.
But we could open a new issue about a feature idea that Jomshir brought up yesterday: Letting users pin their favorite public spaces to some favorite spaces screen, which would help with this case sufficiently in my view.

PR #567, opened 2024-02-03
[FEATURE] second factor captcha

Sekkmer:
require captcha if too many invalid login attempts have been made

image


pandora-client-web/src/components/gameContext/directoryConnectorContextProvider.tsx
```diff
@@ -143,3 +148,76 @@ export function useAuthTokenIsValid(): boolean {
 
 	return authToken != null && isValid;
 }
+
+function SecondFactorDialog() {
+	const directoryConnector = useDirectoryConnector();
+	const [types, setTypes] = React.useState<SecondFactorType[] | null>([]);
+	const [invalid, setInvalid] = React.useState<SecondFactorType[] | null>([]);
+	const [handler, setHandler] = React.useState<null | { resolve: (data: SecondFactorData | PromiseLike<SecondFactorData> | null) => void; }>(null);
```
Jomshir98, line 156: All three of these states are always set at the same time, it looks?
Wouldn't it then be much better and easier to combine them into a single state which is either `null` or object containing valid values of these three?

pandora-server-directory/src/networking/manager_client.ts
```diff
@@ -806,6 +812,31 @@ export const ConnectionManagerClient = new class ConnectionManagerClient impleme
 			connection.sendMessage('somethingChanged', { changes: ['shardList'] });
 		}
 	}
+
+	private async testSecondFactor(data: SecondFactorData | undefined, types: SecondFactorType[]): Promise<null | SecondFactorResponse> {
```
Jomshir98, line 816: This function is never used.

pandora-server-directory/src/networking/manager_client.ts
```diff
@@ -903,3 +934,38 @@ function WithConstantTime<TParams extends unknown[], TReturn extends { /** only
...
+	public async testOptionalCaptcha(data: SecondFactorData | undefined): Promise<null | SecondFactorResponse> {
+		if (!this.isCaptchaRequired()) {
+			return null;
+		}
+		const types: SecondFactorType[] = ['captcha'];
+		if (data == null) {
+			return { result: 'secondFactorRequired', types };
+		}
+		if (data.captcha == null) {
+			this.loginFailed();
```
Jomshir98, line 958: I don't think missing captcha should increase the counter - a valid login can trigger this easily if the client doesn't yet know to include it, which would mean that if this mechanism is triggered, it will keep stacking up the failed logins even if only valid clients continue logging in.

Right now it doesn't matter much as this only triggers if there is _some_ second factor and captcha is the only one, but if we ever introduce another second factor (like TOTP authentication) this might be problem.
I think that second factor being `undefined` and `{}` should be treated the same.

pandora-server-directory/src/config.ts
```diff
@@ -109,6 +109,15 @@ export const EnvParser = CreateEnvParser({
 	HCAPTCHA_SITE_KEY: z.string().default(''),
 
 	//#endregion
+
+	//#region Account Security
+
+	/** Time window for login attempts */
+	LOGIN_ATTEMPT_WINDOW: EnvTimestamp().default('15m'),
+	/** Max failed login attempts before requiring a captcha */
+	LOGIN_ATTEMPT_LIMIT: z.number().int().positive().default(30),
```
Jomshir98, line 118: Directory's `.env.template` file should be updated with these values.

pandora-server-directory/src/networking/manager_client.ts
```diff
@@ -903,3 +934,38 @@ function WithConstantTime<TParams extends unknown[], TReturn extends { /** only
...
+		}
+		if (!await TestCaptcha(data.captcha)) {
+			this.loginFailed();
+			return { result: 'secondFactorInvalid', types, invalid: types, missing: [] };
+		}
+		return null;
+	}
+
+	private isCaptchaRequired(): boolean {
+		return this.invalidAttempts.length >= ENV.LOGIN_ATTEMPT_LIMIT && (this.invalidAttempts[0].timestamp + ENV.LOGIN_ATTEMPT_WINDOW) > Date.now();
```
Jomshir98, line 969: This should not trigger if captcha is not configured at all `(!HCAPTCHA_SECRET_KEY || !HCAPTCHA_SITE_KEY)`.

Jomshir98 (approved):
Looks and works great! Only a few relatively smaller notes.

Jomshir98 (changes requested):
Missclicked the button... rest of the review is still correct :sweat_smile:

pandora-server-directory/src/config.ts
Sekkmer: hmm actually I am not quite sure how much we need the env template
.env is copied only once so no further changes are reflected
if we change any value here users will have an override for the old default
I feel it it would be cleaner if users only ever override what they really want to override

(in any case this PR is not about this so I just added this to for now)

pandora-server-directory/src/networking/manager_client.ts
Sekkmer: ups I left it here

Jomshir98 (approved):
LGTM!

PR #579, opened 2024-02-11
[FEATURE] Full database migration

Jomshir98:
Adds a code for full database migration. This code walks through every single document in all collections and runs it through a supplied Zod schema, updating the document to match (both changed and deleted properties, unlike our current on-the-fly migrations).

The migration is disabled by default and can be activated either in dry-run mode (where the changes are not actually committed to the database) or for actual migration.
These are triggered with env configuration `DATABASE_MIGRATION=dry-run` and `DATABASE_MIGRATION=migrate` respectively.

Note that some collections currently don't have a Zod schema. Those are walked through and ignored. (TODOs have been added)


pandora-server-directory/src/database/mongoDb.ts
```diff
@@ -518,7 +520,137 @@ export default class MongoDatabase implements PandoraDatabase {
...
+					migrationLog.error(`Failed to migrate ${name} document ${documentId.toHexString()}: Update matched count is ${matchedCount}`);
+				}
+			}
+		};
+
+		// Perform migration on all collections
+		await performMigration<DatabaseAccountWithSecure>(
+			'accounts',
+			this._accounts,
+			DatabaseAccountWithSecureSchema,
```
Sekkmer, line 614: my issue with this approach is that it can only do what the load time check already does in some of the cases

what this lacks is true migration when we need to transform the old structure to a new one, like if we want to rename `chatroom` to `space`, or any other more complex type transformation

a more versetile aproach could be to create a collection class, that can do the loading migration all the time, so we do not have to sprinkle it everywhere, and accept 2 optional parameter, old schema, and transformation function

```ts
class ValidatedCollection<TSchema extends ZodTypeAny, TOldSchema extends ZodTypeAny = TSchema> {
	constructor(
		key: string,
		indexes: (IndexDescription & { name: string; })[],
		schema: TSchema,
		oldSchema?: TOldSchema,
		migration?: (old: z.infer<TOldSchema>) => z.infer<TSchema>,
	) {
	}
}
```

Jomshir98: Of course there are many possibilities of approaching this, many more advanced than this.
I maybe wasn't fully clear on the goals of _this_ implementation, those are two:
- Ability to do a "dry run" on the whole database to see what the schema changes actually do on real data - to see that they are not invalidating something that should be valid.
- Ability to sweep through the whole database AoT style instead of only when specific value is loaded to cleanup old, potentially inactive entries.
Yes, we can easily add a pre-/post-parse hooks for manipulating either old data before schema check or data after schema check if the needed migration is not only about schemas.

Sekkmer (approved):

PR #581, opened 2024-02-12
[FEATURE] Join Me invite for spaces

### LINKED ISSUE #571: Improve space joining conditions
The idea was that the behavior would be changed to the following:

You are only able to join an unlisted public rooms (no admin online inside) and private rooms, if:
- you are owner, admin, or on allow list
- ~~it is password protected and you know the password~~ Password mechanic should be removed altogether, see below
- you have a specially crafted valid invite link by an owner or admin (the generic `.../space/join/safgwsuzagasjasdfsafsah` is from my point of view not a valid invite link to join unless the room is listed fully public with no restrictions and anyone can join anyway)
- someone who is currently inside sends you a direct invite via DM that is only valid while they are online inside the space or maybe even additionally expire after 15 mins or so.

Sekkmer:
resolves #571

### DIFF
--- a/pandora-client-web/src/ui/screens/spaceConfiguration/spaceConfiguration.tsx
+++ b/pandora-client-web/src/ui/screens/spaceConfiguration/spaceConfiguration.tsx
@@ -383,7 +383,7 @@ export function SpaceConfiguration({ creation = false }: { creation?: boolean; }
 								<NumberListArea values={ currentConfig.allow } setValues={ (allow) => setModifiedData({ allow }) } readOnly={ !canEdit } invalid={ invalidAllow } />
 							</div>
 						</FieldsetToggle>
-						{ canEdit && currentSpaceId != null && <SpaceInvites spaceId={ currentSpaceId } /> }
+						{ (!creation && currentSpaceId != null) && <SpaceInvites spaceId={ currentSpaceId } isPlayerAdmin={ isPlayerAdmin } /> }
 						<br />
 						{ canEdit && <Button className='fill-x' onClick={ () => UpdateSpace(directoryConnector, modifiedData, () => navigate('/room')) }>Update space</Button> }
 					</div>
@@ -394,7 +394,7 @@ export function SpaceConfiguration({ creation = false }: { creation?: boolean; }
 	);
 }
 
-function SpaceInvites({ spaceId }: { spaceId: SpaceId; }): ReactElement {
+function SpaceInvites({ spaceId, isPlayerAdmin }: { spaceId: SpaceId; isPlayerAdmin: boolean; }): ReactElement {
 	const directoryConnector = useDirectoryConnector();
 	const [invites, setInvites] = useState<readonly SpaceInvite[]>([]);
 	const [showCreation, setShowCreation] = useState(false);
@@ -449,13 +449,13 @@ function SpaceInvites({ spaceId }: { spaceId: SpaceId; }): ReactElement {
 						}
 					</tbody>
 				</table>
-				{ showCreation && <SpaceInviteCreation closeDialog={ () => setShowCreation(false) } addInvite={ addInvite } /> }
+				{ showCreation && <SpaceInviteCreation closeDialog={ () => setShowCreation(false) } addInvite={ addInvite } isPlayerAdmin={ isPlayerAdmin } /> }
 			</Column>
 		</FieldsetToggle>
 	);
 }
 
-function SpaceInviteCreation({ closeDialog, addInvite }: { closeDialog: () => void; addInvite: (invite: SpaceInvite) => void; }): ReactElement {
+function SpaceInviteCreation({ closeDialog, addInvite, isPlayerAdmin }: { closeDialog: () => void; addInvite: (invite: SpaceInvite) => void; isPlayerAdmin: boolean; }): ReactElement {
 	const directoryConnector = useDirectoryConnector();
 	const [allowAccount, setAllowAccount] = useState(false);
 	const [account, setAccount] = useState(0);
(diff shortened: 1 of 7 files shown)


pandora-common/src/space/space.ts
```diff
@@ -61,9 +61,17 @@ export const SpaceInviteSchema = z.object({
 	characterId: CharacterIdSchema.optional(),
 	/** The time when the invite expires */
 	expires: z.number().int().optional(),
+	/** Type of the invite */
+	type: z.enum(['joinMe', 'spaceBound']),
+	/** Creator of the invite */
+	createdBy: z.object({
+		accountId: AccountIdSchema,
+		characterId: CharacterIdSchema,
+		characterName: z.string(),
```
Jomshir98, line 70: I'm not fully sure about storing a name of the character as well. Is it really needed? I think I would prefer only showing ID right now with ability to get it separately for display later.

pandora-server-directory/src/spaces/space.ts
```diff
@@ -414,7 +410,13 @@ export class Space {
 			const invite = this._getValidInvite(character, inviteId);
 			if (!invite)
 				return 'invalidInvite';
-
+			if (invite.type === 'joinMe') {
+				const creator = [...this.characters].find((c) => c.baseInfo.id === invite.createdBy.characterId);
+				if (creator?.assignedClient == null)
+					return 'invalidInvite';
```
Jomshir98, line 416: Use `isOnline()` to check this instead; it is more consistent - not prone to rapid changes on temporary DCs.

pandora-server-directory/src/spaces/space.ts
```diff
@@ -656,6 +700,7 @@ export class Space {
 
 		await this._assignedShard?.update('characters');
 		ConnectionManagerClient.onSpaceListChange();
+		await GetDatabase().updateSpace(this.id, { invites: this._invites }, null);
```
Jomshir98, line 703: Wouldn't it be better if `_cleanupInvites` did this asynchronously if the cleanup actually changed something? It can also be done asynchronously as it doesn't matter if the cleanup fails - it can just be repeated the next time. Only potential problem with that are race conditions...

pandora-common/src/space/space.ts
```diff
@@ -61,9 +61,17 @@ export const SpaceInviteSchema = z.object({
...
+	}),
```
Jomshir98, line 71: These changes cause existing spaces with existing invites to fail to load. The invites in `SpaceDataSchema` should be updated to use `ZodArrayWithInvalidDrop`.

pandora-client-web/src/ui/screens/spaceConfiguration/spaceConfiguration.tsx
```diff
@@ -448,13 +449,13 @@ function SpaceInvites({ spaceId }: { spaceId: SpaceId; }): ReactElement {
...
+				{ showCreation && <SpaceInviteCreation closeDialog={ () => setShowCreation(false) } addInvite={ addInvite } isPlayerAdmin={ isPlayerAdmin } /> }
 			</Column>
 		</FieldsetToggle>
 	);
 }
 
-function SpaceInviteCreation({ closeDialog, addInvite }: { closeDialog: () => void; addInvite: (invite: SpaceInvite) => void; }): ReactElement {
+function SpaceInviteCreation({ closeDialog, addInvite, isPlayerAdmin }: { closeDialog: () => void; addInvite: (invite: SpaceInvite) => void; isPlayerAdmin: boolean; }): ReactElement {
 	const directoryConnector = useDirectoryConnector();
 	const [allowAccount, setAllowAccount] = useState(false);
```
Jomshir98, line 460: Nitpick: This should default to `!isPlayerAdmin`

Jomshir98 (changes requested):
Looks and works good! There are some UX problems, but those can be ironed out over time. Left only a few smaller notes.

pandora-client-web/src/ui/screens/spaceConfiguration/spaceConfiguration.tsx
Sekkmer: why?
the function is only ever called locally (not exported) where this is provided as a variable and not a literal
if we ever need `<SpaceInviteCreation isPlayerAdmin={ false } />` directly only then do I see a reason for this
the function above this `SpaceInvites` is the same

pandora-server-directory/src/spaces/space.ts
Sekkmer: I copied the logic from `hasAdminInside()`, I updating both to use `isOnline()`

pandora-server-directory/src/spaces/space.ts
Sekkmer: actually it could not be used in all the other cases when the clean up changed, since we either required to change or we save later anyway

pandora-client-web/src/ui/screens/spaceConfiguration/spaceConfiguration.tsx
Jomshir98: Because it is not possible for non-admins to create an invite that doesn't make use of account filtering - so if player is not admin, they always need to enable that filter to successfully create an invite that way.
Another alternative I'm thinking about is just removing this menu altogether for non-admins and making this menu only create space-bound invites.

But I don't feel strongly about this (I wrote it is a nitpick in the first place) - feel free to close this thread.

Jomshir98 (approved):

pandora-client-web/src/ui/screens/spaceConfiguration/spaceConfiguration.tsx
Sekkmer: ahh I see, wasn't quite sure about the reasoning, well I just wanted to provide an alternative to `/invite` even if it is a bit raugh~~

in any case I think it is good for now and we can refine it later

PR #587, opened 2024-02-19
[CHANGE] fixes incorrect color picker hue jumps and updates wiki

ClaudiaMia:


Sekkmer (approved):
Stangly the papers do not call for reverse for HSV -> RGB case, but feels logical it is required for both sides

Jomshir98 (approved):

ClaudiaMia:
> Stangly the papers do not call for reverse for HSV -> RGB case, but feels logical it is required for both sides

Yeah, Jomshir looked into the paper and said it is wrong there.

PR #589, opened 2024-02-20
[FEATURE] device enter/leave permission

### LINKED ISSUE #586: Fix that users on the allow list cannot join unlisted spaces
Hi Sekkmer,

there is a bug in your implementation of https://github.com/Project-Pandora-Game/pandora/issues/571:

If a space is either public or private, but it is unlisted (because of no online admin inside), then someone on the allow list cannot join the space, even though they would be allowed, because the space is not listed for them.

Sekkmer:
fixes #586


Jomshir98 (approved):
LGTM!

PR #598, opened 2024-02-29
[FEATURE] limit dms

Sekkmer:
reduce direct message length by x10 to 2500 characters
ensure that only 50 messages are saved at most for a given conversation
store full conversion in a single object in database

### DIFF
--- a/pandora-server-directory/src/account/accountDirectMessages.ts
+++ b/pandora-server-directory/src/account/accountDirectMessages.ts
@@ -1,11 +1,11 @@
 import { createHash } from 'crypto';
-import type { IClientDirectoryArgument, IClientDirectoryPromiseResult, IDirectoryClientArgument, IDirectoryDirectMessage, IDirectoryDirectMessageAccount, IDirectoryDirectMessageInfo } from 'pandora-common';
+import type { IClientDirectoryArgument, IClientDirectoryPromiseResult, IDirectoryClientArgument, IDirectoryDirectMessageAccount, IDirectoryDirectMessageInfo } from 'pandora-common';
 import { GetDatabase } from '../database/databaseProvider';
 import { Account, GetDirectMessageId } from './account';
 import { accountManager } from './accountManager';
 import { DatabaseDirectMessageInfo } from '../database/databaseStructure';
 
-const MESSAGE_LOAD_COUNT = 50;
+const MESSAGE_MAX_COUNT = 50;
 
 let lastMessageTime = 0;
 /** TODO: handle host machine time jumping backwards */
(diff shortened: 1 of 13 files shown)


pandora-server-directory/src/database/mockDb.ts
```diff
@@ -322,42 +322,42 @@ export class MockDatabase implements PandoraDatabase {
...
+			if (message.edited != null)
 				return Promise.resolve(false);
 
-			msg.content = message.content;
-			msg.edited = message.edited;
+			data = {
+				accounts,
+				keyHash,
+				messages: [message],
+			};
```
Jomshir98, line 340: Shouldn't this also actually save the new messaging data?

pandora-server-directory/src/database/mongoDb.ts
```diff
@@ -35,18 +35,305 @@ type MongoDbInit = Readonly<{
...
 export const MONGODB_SERVER_VERSION: string = '6.0.5';
 
+interface DbAutomaticMigration {
+	readonly dryRun: boolean;
+	readonly log: Logger;
+	readonly startTime: number;
+	success: boolean;
+	totalCount: number;
+	changeCount: number;
+}
```
Jomshir98, line 45: There are some rather huge changes to the MongoDB driver, unrelated to the DM changes and not even a separate commit to review them separately. Frankly I don't feel like reviewing this and I feel that database is critical enough component that I don't want to approve it simply on the fact that tests still pass~

pandora-server-directory/src/account/accountDirectMessages.ts
```diff
@@ -1,11 +1,11 @@
 import { createHash } from 'crypto';
-import type { IClientDirectoryArgument, IClientDirectoryPromiseResult, IDirectoryClientArgument, IDirectoryDirectMessage, IDirectoryDirectMessageAccount, IDirectoryDirectMessageInfo } from 'pandora-common';
+import type { IClientDirectoryArgument, IClientDirectoryPromiseResult, IDirectoryClientArgument, IDirectoryDirectMessageAccount, IDirectoryDirectMessageInfo } from 'pandora-common';
 import { GetDatabase } from '../database/databaseProvider';
 import { Account, GetDirectMessageId } from './account';
 import { accountManager } from './accountManager';
 import { DatabaseDirectMessageInfo } from '../database/databaseStructure';
 
-const MESSAGE_LOAD_COUNT = 50;
+const MESSAGE_MAX_COUNT = 50;
```
Jomshir98, line 8: This should be moved to the `inputLimits.ts` file in pandora-common as it is no longer just an implementation detail, but greatly affects the behavior.

pandora-server-directory/src/account/accountDirectMessages.ts
```diff
@@ -99,22 +99,22 @@ export class AccountDirectMessages {
 		}
 		const time = GetNextMessageTime();
 		const accounts = GetDirectMessageId(this._account, target);
-		const message: IDirectoryDirectMessage = {
+		const message = {
```
Jomshir98, line 102: You should always explicitly define type when creating object in-place like this - otherwise TS won't catch if you specify unknown key (as the inferred type will still be assignable where you use it, so it isn't an error)

Jomshir98 (changes requested):
Reviewed everything except MongoDB changes. See the comment there.

pandora-server-directory/src/database/mongoDb.ts
Sekkmer: sure I split that commit so it is more clear what code is just moved and what is new code

pandora-server-directory/src/database/validatedCollection.ts
```diff
@@ -192,3 +224,40 @@ export class ValidatedCollection<T extends Document> {
 		}
 	}
 }
+
+class ValidatedAsyncIter<T extends Document> implements AsyncIterableIterator<T | null> {
```
Jomshir98, line 239: [optional] This could be simplified by making use of async generator functions: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator

Jomshir98 (changes requested):
Looks good after the split! I like the new generic way of doing things for DB; it will save me some trouble with what I'm working on as well.
Added few more review notes; relatively minor ones.

Jomshir98 (approved):
Looks great!

PR #599, opened 2024-03-03
[DEV] Tests coverage collection

Jomshir98:
This PR adds coverage collection for E2E tests (both client and server), and adds few more bits to combine coverage from individual and E2E tests into a single report.
Locally this can be obtained by running `pnpm run ci` in the workspace root (will take a few minutes).
In CI the coverage is collected but not used yet. In the future we could easily setup something like CodeCov to get coverage diff for each PR.

There is a slight slowdown involved in the coverage collection, but I would say relatively minor one.

The coverage collection for client (+common) during E2E testing was easy, but getting it transformed into a reasonable format was not. It is a bit inaccurate due to the minimization the client build does despite the existence of sourcemaps. It took me several hours to get it to report anything else than 0/0 for anything but lines... But it does work, even if the percentage is not accurate and lines might get reported as run even if they weren't, the functions coverage is accurate and branches coverage is accurate for functions that were run at least once.

### DIFF
--- a/.hooks/coverage-collect.js
+++ b/.hooks/coverage-collect.js
@@ -1,0 +1,61 @@
+#!/usr/bin/node
+/* Scripts are run in Node, so don't make use of the logger or ES imports */
+/* eslint-disable no-console, @typescript-eslint/no-var-requires*/
+
+const { constants } = require('fs');
+const { copyFile } = require('fs/promises');
+const path = require('path');
+const { spawnSync } = require('child_process');
+const fs = require('fs');
+const rimraf = require('rimraf');
+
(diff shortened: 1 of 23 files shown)


.hooks/coverage-collect.js
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/node
```
Sekkmer, line 1: should this be `	#!/usr/bin/env node`?

Jomshir98: Good point! Done and updated the other script as well.

Sekkmer (approved):

Issue #578, opened 2024-02-11
Add and communicate a policy for removal of  inactive accounts


ClaudiaMia:
EU law (GDPR) says that sites should delete all inactive users' personal data after a certain amount of time.

We should announce and remove inactive accounts that were not logged in in the last year, every year.

ClaudiaMia:
I will assign this to myself for making a proposal on how to change the privacy policy to include this topic.

ClaudiaMia:
Not needed. See. https://github.com/Project-Pandora-Game/pandora/issues/610

Issue #559, opened 2024-01-30
Allow deletion of direct message (DM) conversations


ClaudiaMia:
The entry of a past conversation with some account on the left should have an X-icon on the right edge of the button (maybe overlapping to the outside, so we do not lose space to show longer account names clearly).

Clicking the button should open a prompt if the user is really sure they want to delete the whole conversation history with account name XY.

This could save us considerable space going forward, as DM conversations can eat up any amount of space and letting users clean up helps us in that regard.

ClaudiaMia:
I think (1) does not need to be available before beta and can be delayed.

I will assign this to myself for making a proposal on how to change the privacy policy to include (2).

Sekkmer:
we may need to add a message like `XZY deleted the conversation at <time>`

otherwise some user may get confused why the conversation is not there any more and may report it as a bug

ClaudiaMia:
About that: I had anyway one question:

Let's say users A and B exchanged 36 messages over the span of several days.
Now user A deletes the conversation.
Will user B now lose this conversation, too in their DM history with user A? I am not sure that is the expected behavior from a user perspective, but then it would mean we would only delete a conversation if the last copy has been deleted. Unless we do not duplicate the same conversation but reference it only. Or maybe deleting a conversation means only deleting your own messages in a conversation for both users, but that would indeed then be confusing without a message as you suggested to be injected instead in the place of every single deleted message of the conversation.

I am not sure we should do something automatically if you block an account. I would leave that up to the user.

Sekkmer:
there is one conversation if we implement delete it deletes for both sides the same, as the issue you made mentions deleting the conversation not the invidual messages

right now you can delete your own messages by sending an empty edit, but the client UI has no support for this

PR #600, opened 2024-03-04
[CHANGE] update privacy policy

ClaudiaMia:
Updates the privacy policy to revision 2 after exactly one year!

Tries to implement (2) of https://github.com/Project-Pandora-Game/pandora/issues/559

and https://github.com/Project-Pandora-Game/pandora/issues/578

Questions I have:

1. Is it enough to update the privacy policy or do we need some terms of use that state this?
2. Do we need some automatic way that deletes messages when they are older than 1 year? Doing it manually sounds very error prone and tedious. I feel we should do that before we update the privacy policy with this PR.

### DIFF
--- a/pandora-client-web/src/components/Eula/privacyPolicy.tsx
+++ b/pandora-client-web/src/components/Eula/privacyPolicy.tsx
@@ -4,7 +4,7 @@ import { ExternalLink } from '../common/link/externalLink';
 // ********************************
 // Update both of these variables whenever you make any changes to EULA and/or Privacy Policy!
 // ********************************
-export const EULA_VERSION = 2;
+export const EULA_VERSION = 1;
 export const EULA_LAST_UPDATED = 'March 04, 2024';
 
 export function PrivacyPolicyContent(): ReactElement {
@@ -77,8 +77,6 @@ export function PrivacyPolicyContent(): ReactElement {
 			<p>We do not share Your personal information with third parties.</p>
 			<h3>How do We store the data?</h3>
 			<p>We will retain the aforementioned data potentially indefinitely so We can fulfill the purposes described in this Privacy Policy; typically for the lifetime of Your Pandora account.</p>
-			<p>If You do not log into Your Pandora account during a full calendar year, We will delete Your Pandora account and all the data stored in it during the following year.</p>
-			<p>In the case of Direct/Private Messages to other users, We will retain the encrypted data for a maximum of one year.</p>
 			<p>The server that stores the data is inside the European Union.</p>
 			<h3>Cookies</h3>
 			<p>When You visit our website, We may collect information from You automatically through cookies or similar technology.</p>
--- a/pandora-client-web/src/components/Eula/privacyPolicy.tsx
+++ b/pandora-client-web/src/components/Eula/privacyPolicy.tsx
@@ -4,8 +4,8 @@ import { ExternalLink } from '../common/link/externalLink';
 // ********************************
 // Update both of these variables whenever you make any changes to EULA and/or Privacy Policy!
 // ********************************
-export const EULA_VERSION = 1;
-export const EULA_LAST_UPDATED = 'March 04, 2023';
+export const EULA_VERSION = 2;
+export const EULA_LAST_UPDATED = 'March 04, 2024';
 
 export function PrivacyPolicyContent(): ReactElement {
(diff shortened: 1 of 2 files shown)


pandora-client-web/src/components/Eula/privacyPolicy.tsx
```diff
@@ -75,8 +75,10 @@ export function PrivacyPolicyContent(): ReactElement {
 				</li>
 			</ul>
 			<p>We do not share Your personal information with third parties.</p>
-			<h3>How do we store the data?</h3>
-			<p>We will retain the aforementioned data potentially indefinitely so we can fulfill the purposes described in this Privacy Policy; typically for the lifetime of Your Pandora account.</p>
+			<h3>How do We store the data?</h3>
+			<p>We will retain the aforementioned data potentially indefinitely so We can fulfill the purposes described in this Privacy Policy; typically for the lifetime of Your Pandora account.</p>
+			<p>If You do not log into Your Pandora account during a full calendar year, We will delete Your Pandora account and all the data stored in it during the following year.</p>
```
Sekkmer, line 80: Should it be only a single year or should we allow more?

pandora-client-web/src/components/Eula/privacyPolicy.tsx
```diff
@@ -75,8 +75,10 @@ export function PrivacyPolicyContent(): ReactElement {
...
+			<p>In the case of Direct/Private Messages to other users, We will retain the encrypted data for a maximum of one year.</p>
```
Sekkmer, line 81: Why is this a different time span (less than a year) compared to the account (more than a year)?

pandora-client-web/src/components/Eula/privacyPolicy.tsx
Sekkmer, line 80: I feel if we add this to ELUA we must have an issue tied to automatic account and direct message deletion

ClaudiaMia: We can allow more, if that is fine with us and the GDPR.

pandora-client-web/src/components/Eula/privacyPolicy.tsx
ClaudiaMia: Wasn't there the rumor that the GDPR demands that it is deleted after 1 year?
I tried looking into it a bit, but did not find any specific section that states this, but some sites seem to interpret it like that.

pandora-client-web/src/components/Eula/privacyPolicy.tsx
ClaudiaMia: I fully agree, but I cannot do that. So that means we leave this PR open for now?

Sekkmer: actually, I think GDPR does not require us to delete user accounts after some inactivity, so it is more like for us to free up space in database

pandora-client-web/src/components/Eula/privacyPolicy.tsx
Sekkmer: no messaging sites that I am aware of delete conversation messages older then 1 year so we can be quite sure this is not a GDPR requirement

pandora-client-web/src/components/Eula/privacyPolicy.tsx
Sekkmer: I think issue is fine, as we just need to make sure we implement the automatic deletion in a year when we need to run the first deletion

ClaudiaMia: Reverted the new text and opened issue:
https://github.com/Project-Pandora-Game/pandora/issues/610

ClaudiaMia: Reverted the new text and opened issue:
https://github.com/Project-Pandora-Game/pandora/issues/610

pandora-client-web/src/components/Eula/privacyPolicy.tsx
ClaudiaMia: Reverted the new text and opened issue:
https://github.com/Project-Pandora-Game/pandora/issues/610

Sekkmer (approved):

PR #609, opened 2024-03-10
[FEATURE] login management

Sekkmer:
- Add password change UI
- Add UI to display current connections
- Bind login token with connections, destroying the token disconnects all associated connections
- Add warning to 'Logout' as it now affects all tabs
- Invalidate login tokens on password reset and change
- Token timeout results in character disconnect
- UI to extend the life of the login token

### DIFF
--- a/pandora-client-web/src/components/settings/securitySettings.tsx
+++ b/pandora-client-web/src/components/settings/securitySettings.tsx
@@ -175,7 +175,7 @@ function SessionExpire({ token }: { token: AuthToken; }): ReactElement {
 	const now = useCurrentTime(60_000);
 
 	return (
-		<p>Your session will expire in { FormatTimeInterval(token.expires - now, 'full') }</p>
+		<p>Your session will expire in { FormatTimeInterval(token.expires - now, 'two-most-significant') }</p>
 	);
 }
 
--- a/pandora-client-web/src/components/settings/securitySettings.tsx
+++ b/pandora-client-web/src/components/settings/securitySettings.tsx
@@ -184,9 +184,12 @@ function ExtendCurrentSessionDialog({ token, hide }: { token: AuthToken; hide: (
 	const [password, setPassword] = React.useState('');
 
 	const [extend, processing] = useAsyncEvent(
-		() => directory.extendAuthToken(password),
-		(result) => {
-			if (result) {
+		async () => {
+			const passwordSha512 = await PrehashPassword(password);
+			return await directory.awaitResponse('extendLoginToken', { passwordSha512 });
+		},
+		({ result }) => {
+			if (result === 'ok') {
 				toast('Session extended', TOAST_OPTIONS_SUCCESS);
 				hide();
 			} else {
--- a/pandora-client-web/src/components/settings/securitySettings.tsx
+++ b/pandora-client-web/src/components/settings/securitySettings.tsx
@@ -5,7 +5,6 @@ import { useCurrentAccount, useDirectoryConnector } from '../gameContext/directo
 import { PrehashPassword } from '../../crypto/helpers';
 import { Form, FormCreateStringValidator, FormField, FormFieldError } from '../common/form/form';
 import { Button } from '../common/button/button';
-import { FieldsetToggle } from '../common/fieldsetToggle/fieldsetToggle';
 import { useAsyncEvent } from '../../common/useEvent';
 import { ModalDialog, useConfirmDialog } from '../dialog/dialog';
 import { useObservable } from '../../observable';
@@ -199,10 +198,14 @@ function ExtendCurrentSessionDialog({ token, hide }: { token: AuthToken; hide: (
 			}
 		},
 	);
+	const onSubmit = React.useCallback((ev: React.FormEvent) => {
+		ev.preventDefault();
+		extend();
+	}, [extend]);
 
 	return (
 		<ModalDialog>
-			<Form dirty={ false } onSubmit={ extend }>
+			<Form dirty={ false } onSubmit={ onSubmit }>
 				<SessionExpire token={ token } />
 				<FormField>
 					<label htmlFor='extend-current-session-password'>Password</label>
--- a/pandora-client-web/src/components/settings/securitySettings.tsx
+++ b/pandora-client-web/src/components/settings/securitySettings.tsx
@@ -209,6 +209,7 @@ function ExtendCurrentSessionDialog({ token, hide }: { token: AuthToken; hide: (
 					<input
 						type='password'
 						id='extend-current-session-password'
+						autoComplete='current-password'
 						value={ password }
 						onChange={ (e) => setPassword(e.target.value) }
 					/>
@@ -281,6 +282,7 @@ function PasswordChange({ account }: { account: IDirectoryAccountInfo; }): React
 					<input
 						type='password'
 						id='password-change-old'
+						autoComplete='current-password'
 						{ ...register('oldPassword', {
 							required: 'Old password is required',
 						}) }
@@ -292,6 +294,7 @@ function PasswordChange({ account }: { account: IDirectoryAccountInfo; }): React
 					<input
 						type='password'
 						id='password-change-new'
+						autoComplete='new-password'
 						{ ...register('newPassword', {
 							required: 'New password is required',
 							validate: FormCreateStringValidator(PasswordSchema, 'password'),
--- a/pandora-client-web/src/components/settings/securitySettings.tsx
+++ b/pandora-client-web/src/components/settings/securitySettings.tsx
@@ -1,5 +1,5 @@
 import React, { ReactElement } from 'react';
-import { AssertNever, PasswordSchema, IDirectoryAccountInfo, IClientDirectoryNormalResult } from 'pandora-common';
+import { AssertNever, FormatTimeInterval, PasswordSchema, IDirectoryAccountInfo, IClientDirectoryNormalResult } from 'pandora-common';
 import { useForm } from 'react-hook-form';
 import { useCurrentAccount, useDirectoryConnector } from '../gameContext/directoryConnectorContextProvider';
 import { PrehashPassword } from '../../crypto/helpers';
@@ -7,7 +7,10 @@ import { Form, FormCreateStringValidator, FormField, FormFieldError } from '../c
 import { Button } from '../common/button/button';
 import { FieldsetToggle } from '../common/fieldsetToggle/fieldsetToggle';
 import { useAsyncEvent } from '../../common/useEvent';
-import { useConfirmDialog } from '../dialog/dialog';
+import { ModalDialog, useConfirmDialog } from '../dialog/dialog';
+import { useObservable } from '../../observable';
+import type { AuthToken } from '../../networking/directoryConnector';
+import { useCurrentTime } from '../../common/useCurrentTime';
 
 export function SecuritySettings(): ReactElement | null {
 	const account = useCurrentAccount();
@@ -59,6 +62,7 @@ function ConnectedClientsList({ connections }: { connections: AccountConnectedCl
 			<thead>
 				<tr>
 					<th>Id</th>
+					<th>Connections</th>
 					<th>Characters</th>
 					<th>Actions</th>
 				</tr>
@@ -74,9 +78,10 @@ function ConnectedClientsList({ connections }: { connections: AccountConnectedCl
 
 function ConnectedClientConnection({ connection }: { connection: AccountConnectedClients[number]; }): ReactElement {
 	const directoryConnector = useDirectoryConnector();
+	const authToken = useObservable(directoryConnector.authToken);
 	const confirm = useConfirmDialog();
 
-	const { loginTokenId, connectedCharacters } = connection;
+	const { loginTokenId, connectionCount, connectedCharacters } = connection;
 
 	const disconnect = React.useCallback(() => {
 		void confirm('Are you sure you want to disconnect this client?').then((confirmed) => {
@@ -86,14 +91,68 @@ function ConnectedClientConnection({ connection }: { connection: AccountConnecte
 		});
 	}, [confirm, directoryConnector, loginTokenId]);
 
+	const isCurrent = authToken?.value.startsWith(loginTokenId);
+
 	return (
-		<tr>
+		<tr className={ isCurrent ? 'current-connection' : undefined }>
 			<td>{ loginTokenId }</td>
+			<td>{ connectionCount }</td>
 			<td>{ connectedCharacters.map(({ id, name }) => `${name} (${id})`).join(', ') }</td>
 			<td>
-				<Button className='slim' onClick={ disconnect }>Disconnect</Button>
+				{
+					isCurrent ? (
+						<>
+							<Button className='slim' onClick={ disconnect }>Logout</Button>
+							<ExtendCurrentSession token={ authToken! } />
+						</>
+					) : (
+						<Button className='slim' onClick={ disconnect }>Disconnect</Button>
+					)
+				}
 			</td>
 		</tr>
+	);
+}
+
+function ExtendCurrentSession({ token }: { token: AuthToken; }): ReactElement {
+	const [show, setShow] = React.useState(false);
+	if (!show)
+		return <Button className='slim' onClick={ () => setShow(true) }>Extend</Button>;
+
+	return (
+		<>
+			<Button className='slim' onClick={ () => setShow(false) }>Extend</Button>
+			<ExtendCurrentSessionDialog token={ token } />
+		</>
+	);
+}
+
+function ExtendCurrentSessionDialog({ token }: { token: AuthToken; }): ReactElement {
+	const directory = useDirectoryConnector();
+	const [password, setPassword] = React.useState('');
+	const now = useCurrentTime(60_000);
+
+	const [extend, processing] = useAsyncEvent(
+		() => directory.extendAuthToken(password),
+		() => setPassword(''),
+	);
+
+	return (
+		<ModalDialog>
+			<Form dirty={ false } onSubmit={ extend }>
+				<p>Your session will expire in { FormatTimeInterval(token.expires - now, 'short') }</p>
+				<FormField>
+					<label htmlFor='extend-current-session-password'>Password</label>
+					<input
+						type='password'
+						id='extend-current-session-password'
+						value={ password }
+						onChange={ (e) => setPassword(e.target.value) }
+					/>
+				</FormField>
+				<Button type='submit' disabled={ processing }>Extend</Button>
+			</Form>
+		</ModalDialog>
 	);
 }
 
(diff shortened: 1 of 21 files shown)


pandora-server-directory/src/account/accountSecure.ts
```diff
@@ -294,20 +353,108 @@ function GenerateSimpleToken(): string {
...
+	}
+
+	public extend(): boolean {
+		if (this.#destroyed)
+			return false;
+
+		this.#expires = Date.now() + TOKEN_TYPES[this.reason].expiration;
+
+		if (this.value.length > ACCOUNT_TOKEN_ID_LENGTH)
+			this.#value = this.getId() + nanoid(32 - ACCOUNT_TOKEN_ID_LENGTH);
```
Jomshir98, line 456: Why does this happen? It in some weird edge-case can regenerate _part_ of the token, which effectively means all existing tokens become invalid, because they differ from this one. That will however not be easily visible, because they will still share the first 32 characters - the visible id. So the token is seemingly valid on both sides, but invalid as on server the rest of the token is different.
That is definitely not a behavior I would expect from `extend` on an existing token.

pandora-client-web/src/components/settings/securitySettings.tsx
```diff
@@ -0,0 +1,273 @@
...
+	);
+
+	const onChange = React.useCallback((open: boolean) => {
+		if (open && connections === null) {
+			void load();
+		}
+	}, [connections, load]);
+
+	return (
+		<FieldsetToggle legend='Connected Clients' open={ false } onChange={ onChange }>
```
Jomshir98, line 49: I'm not sure I like the togglable fieldsets. It is different from all other settings tabs where the fieldsets are simply open all the time. Also in this styling the arrow of each fildset is weirdly misplaced instead of being centered vertically.

pandora-client-web/src/components/settings/securitySettings.tsx
```diff
@@ -0,0 +1,273 @@
...
+			<div>
+				<ConnectedClientsList connections={ connections } />
+				<Button onClick={ load } disabled={ processing }>Refresh</Button>
```
Jomshir98, line 52: The refresh button should be at the top, so user doesn't have to potentially scroll to find it. Also use `Column` instead of simple `div` to get nice spacing between the two elements.

pandora-client-web/src/components/settings/securitySettings.tsx
```diff
@@ -0,0 +1,273 @@
...
+					isCurrent ? (
+						<>
+							<Button className='slim' onClick={ disconnect }>Logout</Button>
+							<ExtendCurrentSession token={ authToken! } />
+						</>
+					) : (
+						<Button className='slim' onClick={ disconnect }>Disconnect</Button>
+					)
+				}
+			</td>
```
Jomshir98, line 114: The buttons look rather squished together... Maybe wrapping them in a `Column` would make it look nicer?

pandora-client-web/src/components/settings/securitySettings.tsx
```diff
@@ -0,0 +1,273 @@
...
+				directoryConnector.sendMessage('logout', { type: 'selected', accountTokenId: loginTokenId });
+			}
+		});
+	}, [confirm, directoryConnector, loginTokenId]);
+
+	const isCurrent = authToken?.value.startsWith(loginTokenId);
+
+	return (
+		<tr className={ isCurrent ? 'current-connection' : undefined }>
+			<td>{ loginTokenId }</td>
```
Jomshir98, line 100: Is the id even a useful information for the user? I feel it mainly just takes up space.
Also it is not really clear which connection is the current one (yes, there is the background, but that seems like common odd-even coloring commonly used for tables until you get to three or even four sessions.
Maybe adding a `(current)` to the id (on a new line under it) of the current session would help with that?

pandora-client-web/src/components/settings/securitySettings.tsx
```diff
@@ -0,0 +1,273 @@
...
+			<td>{ connectionCount }</td>
+			<td>{ connectedCharacters.map(({ id, name }) => `${name} (${id})`).join(', ') }</td>
+			<td>
+				{
+					isCurrent ? (
+						<>
+							<Button className='slim' onClick={ disconnect }>Logout</Button>
+							<ExtendCurrentSession token={ authToken! } />
```
Jomshir98, line 108: I'm not sure how much use is the button to extend the session, especially as the valid time is not even shown anywhere - so you can't see how long you have left.
It is also very obscure thing for anyone not very closely familiar with the concept.
For __user__ it is a button that when your press asks for password and when you enter it... nothing happens?
Yes, there is the text in the dialog itself, but I honestly didn't notice it until I was going through the code - and I did test the interface extensively.

pandora-client-web/src/components/settings/securitySettings.tsx
```diff
@@ -0,0 +1,273 @@
...
+	return (
+		<tr className={ isCurrent ? 'current-connection' : undefined }>
+			<td>{ loginTokenId }</td>
+			<td>{ connectionCount }</td>
+			<td>{ connectedCharacters.map(({ id, name }) => `${name} (${id})`).join(', ') }</td>
+			<td>
+				{
+					isCurrent ? (
+						<>
+							<Button className='slim' onClick={ disconnect }>Logout</Button>
```
Jomshir98, line 107: I'm not fully sure how I feel about the logout button for current session. I feel having only one in the top right might be simpler, this button not adding any value besides having multiple ways to do the same thing.

pandora-client-web/src/components/settings/securitySettings.tsx
```diff
@@ -0,0 +1,273 @@
...
+		<>
+			<ConnectedClients />
+			<PasswordChange account={ account } />
+		</>
+	);
+}
+
+type AccountConnectedClients = IClientDirectoryNormalResult['queryConnections']['connections'];
+
+function ConnectedClients(): ReactElement {
```
Jomshir98, line 33: A message to logout all clients was added to the message definitions, but no button was added for it.
It would be nice to have it added - maybe on the same row as "refresh", just on the right?

pandora-client-web/src/components/settings/securitySettings.tsx
```diff
@@ -0,0 +1,273 @@
...
+				hide();
+			else
+				setPassword('');
+		},
+	);
+
+	return (
+		<ModalDialog>
+			<Form dirty={ false } onSubmit={ extend }>
+				<p>Your session will expire in { FormatTimeInterval(token.expires - now, 'short') }</p>
```
Jomshir98, line 158: Is short time the best format here? Is there any need to keep it limited to the highest unit of time? Short format is primarily meant for places that show quick overview of many items, each having some time, but details should always show most detailed time available.
Also as I said higher I didn't even notice this was here until reading through the code. And the extend password popup is most definitely not a place I would look for this information in.

pandora-client-web/src/components/settings/securitySettings.tsx
```diff
@@ -0,0 +1,273 @@
...
+						id='extend-current-session-password'
+						value={ password }
+						onChange={ (e) => setPassword(e.target.value) }
+					/>
+				</FormField>
+				<br />
+				<Row>
+					<Button onClick={ hide } disabled={ processing }>Cancel</Button>
+					<Button type='submit' disabled={ processing }>Extend</Button>
+				</Row>
```
Jomshir98, line 172: This dialog doesn't work at all. Was it even tested?
Clicking the submit button triggers a page reload, because the submit event does not prevent default.
As there is no code to handle to submit on URL, the page (almost always) reloads quicker than the handler for this runs (even sends to directory thanks to the need to hash things first), so the whole form does nothing...

pandora-client-web/src/components/settings/securitySettings.tsx
```diff
@@ -0,0 +1,273 @@
...
+		<FieldsetToggle legend='Password Change' open={ false }>
+			<Form dirty={ submitCount > 0 } onSubmit={ onSubmit }>
+				<FormField>
+					<label htmlFor='password-change-old'>Old password</label>
+					<input
+						type='password'
+						id='password-change-old'
+						{ ...register('oldPassword', {
+							required: 'Old password is required',
+							validate: FormCreateStringValidator(PasswordSchema, 'password'),
```
Jomshir98, line 237: Current password shouldn't be validated - doing so would make it impossible to change password for anyone who's current password doesn't satisfy our password guidelines... and that is something we definitely don't want.
And yes, my testing account's doesn't satisfy this - that's how I noticed. This would be especially problematic if we ever changed the guidelines for how password should look like.

pandora-client-web/src/networking/socketio_directory_connector.ts
```diff
@@ -383,4 +384,28 @@ export class SocketIODirectoryConnector extends ConnectionBase<IClientDirectory,
...
+		const passwordSha512 = await PrehashPassword(password);
+		const response = await this.awaitResponse('extendLoginToken', { passwordSha512 });
+		switch (response.result) {
+			case 'ok':
+				this._authToken.value = {
+					username,
+					value: response.token,
+					expires: response.expires,
+				};
+				toast('Session extended', TOAST_OPTIONS_SUCCESS);
```
Jomshir98, line 402: Nit: Why is the directory connector responsible for showing the toast with the result? I think it is more sensible to keep UI responsibility to the caller. The directory connector has awfully many responsibilities as it is and they should be more moved out of it instead of new ones being added.

Jomshir98 (changes requested):
Reviewed. See individual comments for details.

pandora-server-directory/src/networking/connection_client.ts
```diff
@@ -69,6 +75,29 @@ export class ClientConnection extends IncomingConnection<IDirectoryClient, IClie
...
+	public bindLoginToken(token: AccountToken): void {
+		Assert(token.reason === AccountTokenReason.LOGIN);
+		token.bind(this);
+		this._loginTokenId = token.getId();
+	}
+
+	public onAccountTokenDestroyed(token: AccountToken): void {
+		if (this._loginTokenId !== token.getId()) {
+			this.logger.warning(`Invalid token destroyed '${this._loginTokenId ?? 'null'}' !== '${token.getId()}'`);
+			return;
```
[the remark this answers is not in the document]
Sekkmer: hmm I put this here as it should never happen, and wanted to make sure like an assert, dropping this check should have no effect as it should always be true

pandora-server-directory/src/account/accountSecure.ts
```diff
@@ -200,40 +236,63 @@ export default class AccountSecure {
...
+	#invalidateToken(reason: AccountTokenReason, isInvalid?: (token: AccountToken) => boolean): void {
+		const tokens: AccountToken[] = [];
+		for (const token of this.#tokens) {
+			if (!token.validate())
+				continue;
+
+			if (token.reason !== reason || (isInvalid != null && !isInvalid(token)))
+				tokens.push(token);
+			else
+				token.onDestroy();
```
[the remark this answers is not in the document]
Sekkmer: I am changing this but I wont remind doing this unless it is part of a lint, since we allow this I am unsure why it would be different in any case

pandora-client-web/src/components/settings/securitySettings.tsx
Sekkmer: issue was that default form action was not prevented, should be good now

pandora-server-directory/src/utility.ts
```diff
@@ -2,3 +2,105 @@
...
+
+	private createExecutionPromise(): Promise<void> {
+		this.shouldScheduleNext = false;
+		this.executingPromise = this.task()
+			.finally(() => {
+				this.executingPromise = null;
+				if (this.shouldScheduleNext) {
+					this.runOnInterval();
+				}
+			});
```
Jomshir98, line 103: This is not enough to safely call `.task()` if it can potentially fail. If the `task` is not async method, but a function that returns promise, it can throw before the promise is returned - the `catch`/`finally` handlers won't be called in that case.
The simplest way to fix this is to make `createExecutionPromise` an `async` method that `await`s the `this.task()` call, wrapping it in a `try`/`finally` block.
This way you can also change the signature of `task` to be `Promise<void> | void` to allow reusing this for simpler timers as well.

pandora-common/src/event.ts
```diff
@@ -6,6 +6,8 @@ export interface ITypedEventEmitter<T extends TypedEvent> {
 	getSubscriber(key: keyof T): (onStoreChange: () => void) => () => void;
 }
 
+export type TypedEventEmitterEvents<T> = T extends ITypedEventEmitter<infer E> ? E : never;
```
Jomshir98, line 9: Nit: Shouldn't this check that the `T` does extend `ITypedEventEmitter<any>` to catch errors early?

pandora-server-directory/src/networking/connection_client.ts
```diff
@@ -63,12 +72,45 @@ export class ClientConnection extends IncomingConnection<IDirectoryClient, IClie
...
+		if (event.tokenDestroyed != null) {
+			Assert(this._loginToken === event.tokenDestroyed);
+			this.onAccountTokenDestroyed();
+		}
+		if (event.extended != null) {
+			Assert(this._loginToken === event.extended);
+			this.sendMessage('loginTokenChanged', {
+				value: event.extended.value,
+				expires: event.extended.expires,
+			});
```
Jomshir98, line 95: This is slightly problematic; imagine the following case:
- You have two tabs open
- In one of the tabs you are extending token, the other tab is on the background
- As the other tab is on the background it can easily DC
- The token refreshes, the current tab does receive token-updating event
- The other tab then reconnects, but it won't receive this and still will have the old data.

I'm not fully sure how to handle this. One possibility I see is having a shared store for the token in `localStorage`, as all open tabs share the value and get update when the saved value changes. In that case only the tab updating the token would store the new data in the store and everyone will see it.

pandora-server-directory/src/networking/connection_client.ts
```diff
@@ -63,12 +72,45 @@ export class ClientConnection extends IncomingConnection<IDirectoryClient, IClie
 			this._account = null;
 		}
 		if (account) {
+			Assert(token?.reason === AccountTokenReason.LOGIN);
 			account.touch();
 			this._account = account;
+			this._loginTokenEventUnsubscribe?.();
```
Jomshir98, line 78: Shouldn't the unsubscribe happen in the higher block about removing the account - i.e. if the call is `setAccount(null)` as well?

pandora-server-directory/src/networking/connection_client.ts
```diff
@@ -63,12 +72,45 @@ export class ClientConnection extends IncomingConnection<IDirectoryClient, IClie
...
+		}
+	}
+
+	private onAccountTokenDestroyed(): void {
+		if (this._account != null) {
+			this.setAccount(null);
+			this.sendConnectionStateUpdate();
+		}
+
+		this.character?.markForDisconnect();
```
Jomshir98, line 105: This will never be called - above you already call `setAccount(null)`, which forcefully detaches the character inside.

pandora-server-directory/src/networking/connection_client.ts
```diff
@@ -63,12 +72,45 @@ export class ClientConnection extends IncomingConnection<IDirectoryClient, IClie
...
+
+		this._loginToken = null;
+		this._loginTokenEventUnsubscribe?.();
+		this._loginTokenEventUnsubscribe = null;
```
Jomshir98, line 109: Again, these feel like they should be done by the `setAccount(null)` call.

pandora-server-directory/src/account/accountSecure.ts
```diff
@@ -294,20 +348,101 @@ function GenerateSimpleToken(): string {
...
+		this.emit('tokenDestroyed', this);
+	}
+
+	public extend(): boolean {
+		if (this.#destroyed)
+			return false;
+
+		const { generate, expiration } = TOKEN_TYPES[this.#reason];
+		this.#value = generate();
+		this.#expires = Date.now() + expiration;
```
Jomshir98, line 442: With this code is there really even a reason to have the "extend" functionality? It seems all this does is effectively invalidate old token and create a brand new one that is then returned.

pandora-client-web/src/components/settings/securitySettings.tsx
```diff
@@ -0,0 +1,327 @@
...
+		handleSubmit,
+		register,
+		trigger,
+	} = useForm<PasswordChangeFormData>({ shouldUseNativeValidation: true, progressive: true });
+
+	React.useEffect(() => {
+		if (invalidPassword) {
+			void trigger();
+		}
+	}, [invalidPassword, trigger]);
```
Jomshir98, line 253: This only re-runs the validation - the `invalidPassword´ variable is actually never displayed anywhere, so there is no feedback at all if user entered invalid current password.

pandora-client-web/src/components/settings/securitySettings.tsx
```diff
@@ -0,0 +1,327 @@
...
+						}) }
+					/>
+					<FormFieldError error={ errors.newPassword } />
+				</FormField>
+				<FormField>
+					<label htmlFor='password-change-new-confirm'>Confirm new password</label>
+					<input
+						type='password'
+						id='password-change-new-confirm'
+						{ ...register('newPasswordConfirm', {
```
Jomshir98, line 313: Per HTML spec the "confirm" field for a new password should _also_ have `autoComplete='new-password'`

Jomshir98 (changes requested):
Done reviewing the changes. Sorry for the threads being detached like this, I did most of the comments on the individual commits.
I do like the new async interval; I would really like to start using it elsewhere too after this gets merged.

pandora-common/src/event.ts
Sekkmer: I tried that but it did not work so that is why I did not add an extends
```
pandora-server-directory build$ pnpm run --silent clean && tsc -p ./tsconfig.json
│ src/networking/connection_client.ts(85,62): error TS2344: Type 'AccountToken' does not satisfy the constraint 'ITypedEventEmitter<any>'.
│   Types of property 'on' are incompatible.
│     Type '<K extends "extended" | "tokenDestroyed">(event: K, listener: (value: { tokenDestroyed: AccountToken; extended: AccountToken; }[K]) => void) => () => void' is not assignable to type '<K extends string | number | symbol>(s: K, listener: (v: any)…
│       Types of parameters 'event' and 's' are incompatible.
│         Type 'K' is not assignable to type '"extended" | "tokenDestroyed"'.
│           Type 'string | number | symbol' is not assignable to type '"extended" | "tokenDestroyed"'.
│             Type 'string' is not assignable to type '"extended" | "tokenDestroyed"'.
```

pandora-server-directory/src/networking/connection_client.ts
Sekkmer: it is already shared:
```ts
private readonly _authToken = BrowserStorage.create<AuthToken | undefined>('authToken', undefined, AuthTokenSchema);
```

pandora-server-directory/src/account/accountSecure.ts
Sekkmer: this is for security reasons to not keep any token valid over a specified time

Jomshir98 (approved):
LGTM!

PR #613, opened 2024-03-15
[REFACTOR] Do not store character information on account

Jomshir98:
This PR moves any character-related data out of database's account object. Instead data for all characters of a single account are gathered by querying the characters collection.

Migration is added to migrate existing data (only current space, really). ~~Do note, that the migration is not technically idempotent as it will prefer old data over new ones if run without `DATABASE_MIGRATION=migrate`, but otherwise it is safe to be run multiple times.~~ Now the migration is fully idempotent as it forces full migration when old data structure is detected.


pandora-server-directory/src/database/mongoDb.ts
```diff
@@ -593,6 +608,79 @@ export default class MongoDatabase implements PandoraDatabase {
...
+						},
+					);
+					Assert(matchedCount === 1);
+
+					charactersToMigrate.delete(migrationInfo.character.id);
+				}
+			},
+		});
+
+		// The character array from the account will be deleted during automatic migration
```
Sekkmer, line 681: there should be a log with total number of migrated acc/chars and we could also log that `charactersToMigrate.size` is 0 at the end

pandora-server-directory/src/database/mongoDb.ts
```diff
@@ -593,6 +608,79 @@ export default class MongoDatabase implements PandoraDatabase {
...
+			migrate: async ({ oldStream }) => {
+				for await (const account of oldStream) {
+					if (Array.isArray(account?.characters)) {
+						for (const character of account.characters) {
+							Assert(!charactersToMigrate.has(character.id));
+
+							charactersToMigrate.set(character.id, {
+								account: account.id,
+								character: {
+									id: character.id,
```
Sekkmer, line 633: account.character will be removed by automatic migration right?

Jomshir98: Yes, if the migration does run in "migrate" mode. I will look into making that automatic if needed~

Sekkmer (approved):

PR #620, opened 2024-03-23
[CHANGE] gate development mode and document space effects

ClaudiaMia:
This PR hides the option to create a space in development mode if you are not having the role of a developer or higher.

Following things need improvement, which are beyond my level of knowledge:

- Make the directory reject a space creation if it includes the effect "development" but the creator is not having the developer role
- Make it so that the space background image or solid color has a large (slightly transparent?) text overlayed "Development mode"

### DIFF
--- a/pandora-client-web/src/ui/screens/spaceConfiguration/spaceConfiguration.tsx
+++ b/pandora-client-web/src/ui/screens/spaceConfiguration/spaceConfiguration.tsx
@@ -292,7 +292,7 @@ export function SpaceConfiguration({ creation = false }: { creation?: boolean; }
 										}
 									} }
 								/>
-								<label htmlFor={ `${idPrefix}-feature-${feature.id}` }>{ ' ' + feature.name }</label>
+								<label htmlFor={ `${idPrefix}-feature-${feature.id}` }> { feature.name }</label>
 							</div>
 						))
 					}
--- a/pandora-client-web/src/ui/screens/spaceConfiguration/spaceConfiguration.tsx
+++ b/pandora-client-web/src/ui/screens/spaceConfiguration/spaceConfiguration.tsx
@@ -22,6 +22,7 @@ import {
 	SpaceInvite,
 	FormatTimeInterval,
 	AssertNever,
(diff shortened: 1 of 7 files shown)


pandora-client-web/src/ui/screens/spaceConfiguration/spaceConfiguration.tsx
```diff
@@ -273,31 +275,30 @@ export function SpaceConfiguration({ creation = false }: { creation?: boolean; }
...
+									} }
+								/>
+								<label htmlFor={ `${idPrefix}-feature-${feature.id}` }>{ ' ' + feature.name }</label>
+							</div>
+						))
+					}
 				</div>
 				{
-					currentConfig.features.includes('development') &&
+					currentConfig.features.includes('development') && isDeveloper &&
```
Sekkmer, line 301: should we also enforce this in server side too, since this only hides the feature and do not prevent anyone from creating a room with development features if they know how

ClaudiaMia: Yes, I wrote so in my comment in the PR description~ *giggles*

Sekkmer: you can add enforcment by checking the feature and the role of the creator, something like this:

in: `pandora-server-directory/src/spaces/spaceManager.ts`

```diff
-       public async createSpace(config: SpaceDirectoryConfig, owners: AccountId[]): Promise<Space | 'failed' | 'spaceOwnershipLimitReached'> {
+       public async createSpace(config: SpaceDirectoryConfig, owners: AccountId[], creator: Account): Promise<Space | 'failed' | 'spaceOwnershipLimitReached'> {
                Assert(owners.length > 0, 'Space must be created with some owners');
 
+               if (config.features.includes('development') && !creator.roles.isAuthorized('developer')) {
+                       return 'failed';
+               }
```

Sekkmer (approved):

PR #622, opened 2024-03-29
[FEATURE] split leg left and right, side order

Sekkmer:
- separate layers for left/right legs and soles
- add user selectable option to for side ordering: weather left or right arm should be in front, same for legs


Sekkmer:
moved the working arm side order to a separate order as leg splitting needs a bit more work to get correct layering

PR #623, opened 2024-03-29
[CHANGE] check owners' role for development on space load

Sekkmer:


Jomshir98:
Hmm... could we maybe just make logic like this part of loading a space from database instead of migration? That way it could stay and we would have logic for removing these settings from spaces that might get abandoned by dev owner at some point (while possibly having another owner) or when someone stops being a dev.

Jomshir98 (approved):
LGTM!

PR #624, opened 2024-03-29
[FEATURE] flag for safemode cooldown bypass

Sekkmer:
This change all previously created development spaces as they will no longer have the bypass safemode cooldown feature

New spaces will need to enable this at creation time

### DIFF
--- a/pandora-common/src/space/space.ts
+++ b/pandora-common/src/space/space.ts
@@ -30,6 +30,7 @@ export type SpaceFeature = z.infer<typeof SpaceFeatureSchema>;
 
 export type ActionSpaceContext = {
 	features: readonly SpaceFeature[];
+	development: Readonly<SpaceDirectoryConfig['development']>;
 	isAdmin(account: AccountId): boolean;
 };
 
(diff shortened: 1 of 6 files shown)


pandora-common/src/space/space.ts
```diff
@@ -83,6 +84,8 @@ export const SpaceDirectoryConfigSchema = SpaceBaseInfoSchema.extend({
 		shardId: z.string().optional(),
 		/** Automatically grants admin to every developer on enter */
 		autoAdmin: z.boolean().optional(),
+		/** Disable safemode cooldown for everyone inside the space */
+		disableCooldown: z.boolean().optional(),
```
Jomshir98, line 88: Nitpick: Maybe a bit more explicit name `disableSafemodeCooldown`, in case we ever get other cooldowns (which is likely, for example apply/modify/remove item cooldown)

pandora-common/src/space/space.ts
```diff
@@ -30,6 +30,7 @@ export type SpaceFeature = z.infer<typeof SpaceFeatureSchema>;
 
 export type ActionSpaceContext = {
 	features: readonly SpaceFeature[];
+	development: Readonly<SpaceDirectoryConfig['development']>;
```
Jomshir98, line 33: Might be worth pulling the development config into its own type by now?

Jomshir98 (approved):
Looks good to me, just one minor note.

Jomshir98 (approved):

PR #628, opened 2024-04-01
[FIX] move character name input to prevent autocomplete

Sekkmer:


Jomshir98 (approved):
I pushed two more commits on top of your changes. Looks good to me now! Feel free to merge if you are happy with my changes, or shout at me if not~ :smile:

Issue #639, opened 2024-04-06
Fix online status for both characters and accounts


ClaudiaMia:
It sometimes happens that the logged in characters of your online contacts do not show and possibly also the account online status is not always accurate.

PR #646, opened 2024-04-08
[FIX] correctly set updating guard to default state

Sekkmer:
this may or may not fix the inconsistent status in contacts (#639) but it should certainly help


Jomshir98 (approved):
Good catch!

PR #647, opened 2024-04-08
[CHANGE] Several database migrations

Jomshir98:
This PR adds several migrations to the database for old data (we really want to migrate before beta to be safe). All added migrations are idempotent.
- Store account/character id counters explicitly in a config document instead of generating them by scraping account and character collections
- Change the character collection to use string CharacterId instead of translating it to number and back
- Rename `chatrooms` collection to `spaces`
- Change space id format from r/abc to s/abc
- Minor cleanup of no longer needed utilities after these migrations

### DIFF
--- a/pandora-server-directory/src/database/mongoDb.ts
+++ b/pandora-server-directory/src/database/mongoDb.ts
@@ -378,15 +378,8 @@ export default class MongoDatabase implements PandoraDatabase {
 	public async getCharactersForAccount(accountId: number): Promise<DatabaseCharacterSelfInfo[]> {
 		const result: DatabaseCharacterSelfInfo[] = await this._characters
 			.find({ accountId })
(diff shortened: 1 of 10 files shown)


pandora-server-directory/src/database/mongoDb.ts
```diff
@@ -380,10 +378,10 @@ export default class MongoDatabase implements PandoraDatabase {
 	public async getCharactersForAccount(accountId: number): Promise<DatabaseCharacterSelfInfo[]> {
 		const result: DatabaseCharacterSelfInfo[] = await this._characters
 			.find({ accountId })
-			.project<Pick<DatabaseCharacterData, 'id' | 'name' | 'preview' | 'currentSpace' | 'inCreation'>>({ id: 1, name: 1, preview: 1, currentSpace: 1, inCreation: 1 })
+			.project<Pick<ICharacterData, 'id' | 'name' | 'preview' | 'currentSpace' | 'inCreation'>>({ id: 1, name: 1, preview: 1, currentSpace: 1, inCreation: 1 })
 			.toArray()
 			.then((p) => p.map((c): DatabaseCharacterSelfInfo => ({
```
Sekkmer, line 383: do we need this map, it feels like a noop now

PR #650, opened 2024-04-08
[FEATURE] arm side order

Sekkmer:
add user configurable side order for arms, structure added for upper arm in case we want to split arms more in the future

this PR split the working arm order code changes from #622

### DIFF
--- a/pandora-client-web/src/graphics/def.ts
+++ b/pandora-client-web/src/graphics/def.ts
@@ -39,9 +39,13 @@ export type ComputedLayerPriority = Exclude<LayerPriority, DoubleOrdered>
 
 type LeftRight = 'LEFT' | 'RIGHT';
 
-export function useComputedLayerPriority({ view }: Immutable<AppearancePose>): readonly ComputedLayerPriority[] {
+export function useComputedLayerPriority({ view, armsOrder }: Immutable<AppearancePose>): readonly ComputedLayerPriority[] {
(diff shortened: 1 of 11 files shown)


pandora-client-web/src/graphics/def.ts
```diff
@@ -26,42 +37,61 @@ export type ComputedLayerPriority = Exclude<LayerPriority, DoubleOrdered>
...
-	throw new Error('COMPUTED_LAYER_ORDERING not valid');
+type LeftRight = 'LEFT' | 'RIGHT';
+
+export function useComputedLayerPriority({ view, armsOrder }: Immutable<AppearancePose>): readonly ComputedLayerPriority[] {
+	return useMemo(() => {
+		const arms: [LeftRight, LeftRight] = ['LEFT', 'RIGHT'];
+
+		if (armsOrder.upper === 'right') {
+			arms.reverse();
+		}
```
Jomshir98, line 48: The first arm defined in `arms` will be the one more in the back. That means these should be swapped as "left first" to me means that the left arm is more in the front.

Jomshir98 (approved):
Looks good to me! Only one minor note on the ordering.
Also tested with some lock-arms assets (e.g. wrist cuffs). Those will need additional pose limits to get the arm ordering right as they expect it (because they use arm layers to do bits that reach between the arms), but that is minor and to be fixed by assets, not this PR.

PR #651, opened 2024-04-10
[CHANGE] renamed errFull to spaceFull and improved permission dialog

ClaudiaMia:
Renamed a string value, because as a user, it is very confusing when you try to join a space and get an error toast about `errFull`.

These messages were clearly not made end user friendly, but I at least wanted to make it a little bit better by calling it `spaceFull` instead of `errFull`.

Also improves the permission dialog clarity, as it was reported to be an issue and I can see why.

### DIFF
--- a/pandora-client-web/src/components/settings/permissionsSettings.tsx
+++ b/pandora-client-web/src/components/settings/permissionsSettings.tsx
@@ -528,8 +528,8 @@ function PermissionPromptDialog({ prompt: { source, requiredPermissions, message
 				}
 			</Column>
 			<Row padding='medium' alignX='space-between' alignY='center'>
-				<Button onClick={ dismiss }>Dismiss</Button>
-				<Button onClick={ acceptAll } disabled={ !allowAccept } className='fadeDisabled'>Allow all</Button>
+				<Button onClick={ dismiss }>Deny unchosen once</Button>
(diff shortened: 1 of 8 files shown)


Sekkmer (changes requested):
I think adding a correct fix is better, as in implement a switch case for the response and write a custom text for the user

https://github.com/Project-Pandora-Game/pandora/blob/ed5e5cca28bdf99648b2d5ab0ba2e3305e17f46f/pandora-client-web/src/ui/screens/spacesSearch/spacesSearch.tsx#L261

ClaudiaMia:
> I think adding a correct fix is better, as in implement a switch case for the response and write a custom text for the user
> 
> https://github.com/Project-Pandora-Game/pandora/blob/ed5e5cca28bdf99648b2d5ab0ba2e3305e17f46f/pandora-client-web/src/ui/screens/spacesSearch/spacesSearch.tsx#L261



> I think adding a correct fix is better, as in implement a switch case for the response and write a custom text for the user
> 
> https://github.com/Project-Pandora-Game/pandora/blob/ed5e5cca28bdf99648b2d5ab0ba2e3305e17f46f/pandora-client-web/src/ui/screens/spacesSearch/spacesSearch.tsx#L261

Sure, but I want to get a quick fix in. We can make a new issue with your suggestion.

Sekkmer (approved):

PR #652, opened 2024-04-10
[FEATURE] Adds three simple commands to change the leg pose

SandrinePDR:
Adds commands for sitting (sit), kneeling (kneel) or standing (up) without using the pose menu


pandora-client-web/src/ui/components/chat/commands.ts
```diff
@@ -307,4 +308,62 @@ export const COMMANDS: readonly IClientCommand<ICommandExecutionContextClient>[]
 				return true;
 			}),
 	},
+	// Commands to change the player's pose
```
Sekkmer, line 311: there is also a turn command it may worth moving it down here

also you could use:
```ts
//#region Commands to change the player's pose
//#endregion
```

that way VSCode allow a nice collapse option

SandrinePDR: Good idea. I'll use it when I add the more sophisticated pose commands.

Sekkmer (approved):

PR #662, opened 2024-04-11
[FIX] directory action message target property

### LINKED ISSUE #654: Bug: room kick shows wrong target in message (source is shown instead)
`Sekkmer (c1) has been kicked by Sekkmer (c1).`

Sekkmer:
fixes #654

add `satisfies` to prevent further errors

test:
image


ClaudiaMia (approved):
LGTM!

PR #663, opened 2024-04-11
[FIX] color input mouse up should be handled even outside

Sekkmer:
capture felt the best solution, this slightly change the behavior dragging the pointer out will still continue the change (I feel this is a bit better)

tried adding `onPointerOut` but it wasn't stable so capture is over all the better

(tested with multiple color inputs at the same time could not get it to stuck or broken in any way)


ClaudiaMia (approved):
Tested and LGTM!

PR #671, opened 2024-04-13
[CHANGE] Replace arrows with /u-code

SandrinePDR:
Change the arrows in autocomplete messages to their unicode codes fo…r better encoding


pandora-common/src/commands/executor.ts
```diff
@@ -80,7 +80,7 @@ export class CommandRunnerExecutor<
 
 	public autocomplete(): CommandAutocompleteResult {
 		return this.options.restArgName ? {
-			header: `🡆<${this.options.restArgName}>🡄`,
+			header: `⮕<${this.options.restArgName}>⬅`,
```
ClaudiaMia, line 83: I do not really like those new arrows, they are way too thin and look more like lines in the font size we use them in.

The purpose of these signs is to tell you at a glance the position in the command variable chain you are currently at if there are several ones.

Since the original one does not work on certain devices, I feel we should find something totally new to indicate the purpose.

How about `►<option>◄` ? Or something totally different and new~

ClaudiaMia (changes requested):

pandora-common/src/commands/executor.ts
SandrinePDR: That were the closest one I could find in regards to what I was told it looked like.
image

The question would be if we need the arrows (or other symbols) in the first place?

Sekkmer: we could also try custom SVGs to make it work on any device and get the exact style we want

(I don't have a strong preference over `🡆`, `⮕`, `►`, to me all looks fine as long as it is better for most devices)

ClaudiaMia: > The question would be if we need the arrows (or other symbols) in the first place?

I mean nina specifically added that as a feature, so it maybe would be a shame to remove it as it could help someone?
Especially if we ever have more complex commands with 5 or more config values.

ClaudiaMia (approved):
Looking good to me!

Jomshir98 (changes requested):
I like these arrows, but how about actually using a u-code like the title says for better compatibility?
That means replacing the characters:
- `▶` with `\u25B6`
- `◀` with `\u25C0`

Jomshir98 (approved):

PR #688, opened 2024-04-13
[FIX] Fix DM chat scrollbar missing

### LINKED ISSUE #682: Investigate missing scrollbar in the DM chat view
See screenshot: https://discord.com/channels/872284471611760720/872568378190086174/1228107740778594315

Loon8128:
Fixes #682

The current situation is that the scroll bar never appears, and DM elements all shrink to fit (with flexbox's not-helpful "hover over message to make it larger"). The root container is still flex (and so shares with the text entry area), but the messages is a block which lets inner messages stack + overflow to a scrollbar properly.

I tried various different solutions, the ultimate problem is that "fixed size elements, scroll if overflow" doesn't seem to be an optimal use case for a flex container, which seems to have two main options: (1) min size elements (i.e. using `flex-shrink: 0`) that don't scroll, or (2) elements that shrink to fit (current behavior)

Even when shrinking the viewport this seemed to maintain behavior, and enabled scrolling in the list. The chat entry area is fine in both a larger and smaller viewport.

## Before

before 1
before 2

## After

after 1
after 2

### DIFF
--- a/pandora-client-web/src/components/directMessage/directMessage.scss
+++ b/pandora-client-web/src/components/directMessage/directMessage.scss
@@ -7,7 +7,6 @@
 	background-color: $grey-darker;
 
 	.direct-message-list {
-		display: flex;
 		flex-flow: column;
 		flex: 1;
 		overflow-x: hidden;


Jomshir98 (changes requested):
This issue is caused by a specific combination of things (including Chrome browser; on Firefox it doesn't happen, which is why I didn't encounter it until now).
The actual problematic part isn't the flex display - that is correct. (also removing only it definitely wouldn't be enough if you would want to clean things up properly)
The root source of the problem is our fix for Zalgo text (text that overflows bounds of the message by using, for example, many combining characters) in combination with Chrome. It appears that setting `overflow: clip` or `overflow: hidden` on Chrome makes it compute the bounding box for flex children differently, which causes this issue.
A better fix would be changing the direct message rendering to work like the chat rendering, where this currently works without problems.

Jomshir98:
This PR should be superseded by #697.
Please let me know if the variant in that PR fixes the issue on your end as well!

Loon8128:
Yes, that PR does also resolve the issue for me. Thanks for the more informed look at it!

PR #692, opened 2024-04-13
[FIX] Don't insert a space for `/me` messages that start with `'s`

### LINKED ISSUE #683: Using a an emote starting with "'s" produces an undesired space
`*'s places a space between the name and the 's.`

Loon8128:
Fixes #683 

I'm not sure _at all_ what I'm doing here, so do let me know if there is a better way or place to do this, I'm slowly picking my way around this codebase. This should implement the rendering of said messages only on receiving on the client, and only for `'s`

### Questions

- Is this the right place for this formatting?
- Are there other situations in which the space should be elided? (Right now it's only messages prefixed with precisely `'s`)
- All of `*'s thing` and `* 's thing`, and `/me 's thing` produce `Name's thing` in chat, however note that `/me's` is invalid (as it fails to parse as a command). Intentional?

Testing various weird chat messages (including those with bold/italic, and multiline emote / non-emote parts all seems to be working as intended (although that's just the chat parser being very robust, nothing to do with this change)

### DIFF
--- a/pandora-client-web/src/ui/components/chat/chat.tsx
+++ b/pandora-client-web/src/ui/components/chat/chat.tsx
@@ -232,7 +232,7 @@ function DisplayName({ message, color }: { message: IChatMessageChat; color: str
 			default:
 				return ['', ''];
 		}
-	}, [message.type]);
+	}, [message.type, message.parts]);
 
 	const onClick = useCallback((event: React.MouseEvent<HTMLSpanElement>) => {
 		event.stopPropagation();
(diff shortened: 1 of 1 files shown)


Sekkmer (approved):
> Is this the right place for this formatting?
- yes

> Are there other situations in which the space should be elided? (Right now it's only messages prefixed with precisely 's)
- maybe, as some names could end with `s` in which case you would use just an `'` without s
(we do not handle this anywhere right now so we don't need to support it here either)

> All of *'s thing and * 's thing, and /me 's thing produce Name's thing in chat, however note that /me's is invalid (as it fails to parse as a command). Intentional?
- yes, `/me's` would be a new command with the `"me's"` as a key, I don't thing we should add this and the current `"me 's"` is enough

PR #694, opened 2024-04-13
[CHANGE] Unify style of profile editor and display

lina-cat:
Uses the same visual style and fonts for profile and profile editor.
This allows for the creation of ASCII art, visually consistent indentation ect.

### DIFF
--- a/pandora-client-web/src/components/profileScreens/profileScreens.scss
+++ b/pandora-client-web/src/components/profileScreens/profileScreens.scss
@@ -52,5 +52,11 @@
 		background-color: $grey-lightest;
 		padding: 0.5em;
 		border: 1px solid black;
+		font-family: Arial, Helvetica, sans-serif;
+		white-space: pre-wrap;
+	}
+
+	.profileEdit {
+		border: 2px dashed $red-dark;
 	}
 }
--- a/pandora-client-web/src/components/profileScreens/accountProfile.tsx
+++ b/pandora-client-web/src/components/profileScreens/accountProfile.tsx
@@ -122,7 +122,7 @@ function AccountProfileDescription({ profileDescription, allowEdit }: { profileD
 					</Button>
 				</Row>
 				<textarea
-					className='flex-1'
+					className='flex-1 profileDescriptionContent profileEdit'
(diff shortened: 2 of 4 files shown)


pandora-client-web/src/components/profileScreens/profileScreens.scss
```diff
@@ -52,5 +52,11 @@
 		background-color: $grey-lightest;
 		padding: 0.5em;
 		border: 1px solid black;
+		font-family: Arial, Helvetica, sans-serif;
+		white-space: pre-wrap;
+	}
+
+	.profileEdit {
+		border: 2px dashed $red-dark;
```
ClaudiaMia, line 60: I think this is absolutely fine as a quick improvement of this feature.

I just wondered about what Ellie said, that typically there would also be a banner above the input field or even higher on the page in yellow or red that warns about you being in edit mode.

Because if you write for half an hour on your profile and then make the silly mistake of assuming that it autosaved and you can press back or otherwise navigate away, you will be devastated!

A potentially most simple solution to deal with that is to give the read mode the same background color as the chat, so:

```
		color: #000;
		background-color: $grey-lightest;
```

and then in the `.profileEdit`

```
		border: 2px dashed $red-dark;
		color: #000;
		background-color: $grey-lightest;
```

But those are just some thoughts of mine. I just feel in the long term we need to address the risk of losing your changes even further.

ClaudiaMia (approved):
Tested that it works well.

Needs 2nd review.

Sekkmer (approved):
looks good to me too

@ClaudiaMia feel free to merge this after closing the comment

pandora-client-web/src/components/profileScreens/profileScreens.scss
lina-cat: for future solutions of this, I would anyway recommend to save the "in progress" edit at least in the browser so you can recover it even after you close the window (plus a button to reset to current profile) but that would really be out of scope here

lina-cat: pushed changes, so another quick review would be appreciated before merge

PR #695, opened 2024-04-13
[FEATURE] Add setting to increase the font size of the name under a character on the canvas

### LINKED ISSUE #691: Add setting to increase the font size of the name under a character on the canvas
We already allow increase the chat font size in the setting, but there you should also be able to increase the font size of the name under a character on the canvas

Angela-BC:
closes #691 

Adds a font size select drop down to settings - interface - chatroom UI.
Renders the name on the canvas with different sizes depending on the selected font size.


pandora-client-web/src/graphics/room/roomCharacter.tsx
```diff
@@ -301,6 +301,16 @@ const RoomCharacterDisplay = React.forwardRef(function RoomCharacterDisplay({
 
 	const showName = useObservable(SettingDisplayCharacterName);
 
+	// NOTE: Can this be done in a cleaner way?
+	let fontScale: number = 1.0;
+	switch (interfaceChatroomCharacterNameFontSize) {
```
ClaudiaMia, line 306: I feel it might still not be big enough for some people.

Maybe we should actually increase the default a bit from the start, considering many may have smaller screens?

Then the factors could be more like: 0.6 / 1.0 / 1.4 (new default) /  1.8 / 2.2

ClaudiaMia (approved):
Aside from my optional comment, it looks good to me and I tested that it works well.

Needs second review.

pandora-client-web/src/graphics/room/roomCharacter.tsx
Angela-BC: Committed a new version with the proposed values. 
On small screens it is still not very large.

Sekkmer (approved):
Looks good to me

(NOTE part is good enough, we can refine it later if needed)

PR #696, opened 2024-04-13
[CHANGE] Standardize ColorInput Component

### LINKED ISSUE #669: Standardize Color Input Style
Right now the color picker for choosing your account and character name colors use the default HTML <input type="color"> tag for handling color selection, causing it to use whatever your browser has built in whereas editing the color for an asset in the wardrobe brings up a specially designed component instead. 

I feel as though these should be standardized, likely by turning the wardrobe color input into a generic input component if it's not already, so it can be used globally, thereby allowing it to also be used for setting the character/account name colors as well as any other spots where it might be needed in the future.

Madmanmayson:
Closes #669 

This changes the input style for the basic `ColorInput` component to use the internal `ColorEditor` created for the `ColorInputRGBA` component. Because of this `ColorInput` has a new parameter required `title` which was retroactively included in the 3 locations the component was already used. 
Two magic numbers for representing max alpha inside the `colorinput.tsx` file were also converted to use the static global variable `Color.maxAlpha` instead of a hardcoded `255` like all other references in the file.

I chose to keep the `ColorInput` component instead of combining and converting them all into `ColorInputRGBA` across all use cases as I feel there could be good semantic reasons for having both as not all color fields should support alpha channels and it might become less clear if both use cases shared the same component. This could be changed upon request though to reduce code reusage.

### DIFF
--- a/pandora-client-web/src/components/common/colorInput/colorInput.tsx
+++ b/pandora-client-web/src/components/common/colorInput/colorInput.tsx
@@ -12,7 +12,7 @@ export function ColorInput({
 	throttle = 0,
 	disabled = false,
 	hideTextInput = false,
-	inputColorTitle,
+	title,
 }: {
 	initialValue: HexColorString;
 	resetValue?: HexColorString;
@@ -20,9 +20,10 @@ export function ColorInput({
 	throttle?: number;
 	disabled?: boolean;
 	hideTextInput?: boolean;
-	inputColorTitle?: string;
+	title: string;
 }): ReactElement {
 	const [value, setInput] = useState<HexColorString>(initialValue.toUpperCase() as HexColorString);
+	const [showEditor, setShowEditor] = useState(false);
 
 	const onChangeCaller = useCallback((color: HexColorString) => onChange?.(color), [onChange]);
 	const onChangeCallerThrottled = useMemo(() => throttle <= 0 ? onChangeCaller : _.throttle(onChangeCaller, throttle), [onChangeCaller, throttle]);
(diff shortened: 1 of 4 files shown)


Sekkmer:
we could keep the difference for outside while having less duplication by:

```ts
type ColorInputProps<THexColor extends `#${string}`> = {
	initialValue: THexColor;
	resetValue?: THexColor;
	onChange?: (value: THexColor) => void;
	throttle?: number;
	disabled?: boolean;
	minAlpha?: number;
	hideTextInput?: boolean;
	title: string;
};

export function ColorInput(props: ColorInputProps<HexColorString>): ReactElement {
	return (
		<ColorInputRGBA { ...props } minAlpha={ Color.maxAlpha } />
	);
}

type ColorInputRGBAProps = ColorInputProps<HexRGBAColorString> & {
	initialValue: HexRGBAColorString;
	resetValue?: HexRGBAColorString;
	onChange?: (value: HexRGBAColorString) => void;
	minAlpha: number;
};

export function ColorInputRGBA({
	initialValue, resetValue, onChange, throttle = 0, disabled = false, minAlpha = 255, title,
}: ColorInputRGBAProps): ReactElement {
```

and in `changeCallback` we can use the schema based on the `minAlpha === Color.maxAlpha`

Madmanmayson:
Ah right that is a solution I overlooked from my initial pass of the code. It seems like now would be a good time to implement it too since the PR is already about bringing the two ColorInput components in line with each other.

I'll work on making these changes in a few hours.

pandora-client-web/src/components/common/colorInput/colorInput.tsx
```diff
@@ -5,77 +5,29 @@ import { Button } from '../button/button';
...
-		setShowEditor(true);
-	}, [setShowEditor]);
-
-	const onInputChange = (ev: ChangeEvent<HTMLInputElement>) => changeCallback(ev.target.value);
+};
 
+export function ColorInput(props: ColorInputProps<HexColorString>): ReactElement {
 	return (
-		<>
-			{ !hideTextInput && <input type='text' value={ value } onChange={ onInputChange } disabled={ disabled } maxLength={ 7 } /> }
```
Sekkmer, line 43: `hideTextInput` option was not implemented for `ColorInputRGBA` and this change removes it from `ColorInput` too

`hideTextInput` should be added to `ColorInputRGBA`

Madmanmayson: All props should now be properly handled as they were before inside `ColorInputRGBA`

Sekkmer (approved):

PR #700, opened 2024-04-14
[FIX] _ should be displayed as is in character name and assets

### LINKED ISSUE #664: Bug: character names with `_` parsed incorrectly in server messages
currently all server messages are passed through style parser but since character names may contain `_` some character can appear incorrect like  _italic_ or __bold__

Sekkmer:
fixes #664

test:
image


ClaudiaMia (approved):
Tested that it works and I also feel that the code makes sense.

It is nice that you also added the escaping to asset names, although I do not think I would approve assets with a name with "_"

But when we have the custom names, this could happen I guess~

PR #705, opened 2024-04-15
[FEATURE] pose preset storage

Sekkmer:
- store up to 20 presets / account
- import export


ClaudiaMia:
I just tested it and it really is very nice! I like it~

Three improvement proposals for the future:
1. An indicator of how many slots you currently use from the maximum amount of slots you have.
2. A way to import several multiple pose presets at once, so that I can exchange whole play sets. So that the importer does not look for one valid string, but looks if there are more afterwards and tries to add as many as possible.
3. Maybe a button to delete all poses at once with an are you sure dialog.

PR #707, opened 2024-04-15
[CHANGE] Change color scheme of profile editor

lina-cat:
The new color scheme is closer to the one used in other places e.g. chat and allows for easier distinguishing of edit and display mode.

__This is a change based on a conversation on https://github.com/Project-Pandora-Game/pandora/pull/694 that did not make it into  that pullrequest before merge__


ClaudiaMia (approved):
Looking good to me!

Sekkmer (approved):

PR #713, opened 2024-04-17
[ADD] show device modules in wearable part

Sekkmer:


Sekkmer:
@ClaudiaMia I fixed the issue, it ended up a bit more complex then I thought it would be, when you test it please click true the wardrobe to see if anything weird or different as the focused item handling is completely replaced

purpleharmony:
Just a note: the new changes clash with #715 . When this is merged, [this line](https://github.com/Project-Pandora-Game/pandora/blob/7a86e173d32badd490d0463f409a3e053b392219/pandora-client-web/src/components/wardrobe/wardrobeItems.tsx#L111) will need to be updated to use the new focuser, depending on the order in which the two are merged.

ClaudiaMia (approved):
Hi Sekkmer, I tested it in as many ways as I could think of.

In general, it works very fine, I just noticed two things:

1. When you add a lock into a lock slot, the UI does not go back to the focused item view as you would expect it afterwards, but it goes back to the wardrobe with no item selected, which requires you to search for the item again and click it, so you lock the lock. Not sure if this is a behavior of this PR or not. It is the same pattern in the room inventory view.

2. This new feature enables you to move an item out room devices with storage modules directly onto your character body (and vice versa), without having to first drop it to the room inventory. That this is working now as a side effect is very amazing.

Sekkmer:
fixes #721

PR #715, opened 2024-04-18
[ADD] a shortcut from the device context menu to the inventory view

purpleharmony:
This adds a shortcut from room devices' context menus to their menus in the room inventory view.

It is fully working as far as I know, but this is my first contribution to pandora and first time working with react, so there are a few design decisions and technical points I'm unsure of and I'd like to get opinions on:

**Design decision:**
- I turned the menu's header with the device name into the button to reach the inventory in order to save space, but maybe you'd prefer having a different button lower in the menu?

**Technical points:**
- Is the location state the right tool to pass the information to the inventory page? Or would it be better to use something else, e.g an optional route parameter.
- If so, is it the right way to clear the location state after use?
- I wanted the list of room items to scroll to the item we're focusing. I added a state in the wardrobe context for that. Is there a better way to pass the information?

### DIFF
--- a/pandora-client-web/src/components/wardrobe/views/wardrobeItemView.tsx
+++ b/pandora-client-web/src/components/wardrobe/views/wardrobeItemView.tsx
@@ -14,7 +14,7 @@ import {
 	ITEM_LIMIT_ROOM_INVENTORY,
 	AppearanceItemsCalculateTotalCount,
 } from 'pandora-common';
-import React, { ReactElement, useEffect, useMemo } from 'react';
+import React, { ReactElement, useEffect, useMemo, useRef } from 'react';
 import { useObservable } from '../../../observable';
 import { isEqual } from 'lodash';
 import { IItemModule } from 'pandora-common/dist/assets/modules/common';
@@ -252,9 +252,17 @@ function InventoryItemViewList({ item, selected = false, setFocus, singleItemCon
 	setFocus?: (newFocus: WardrobeFocus) => void;
 	singleItemContainer?: boolean;
 }): ReactElement {
-	const { targetSelector, target, extraItemActions, heldItem, setHeldItem } = useWardrobeContext();
+	const { targetSelector, target, extraItemActions, heldItem, setHeldItem, scrollToItem, setScrollToItem } = useWardrobeContext();
 	const wornItem = useWardrobeTargetItem(target, item);
 	const extraActions = useObservable(extraItemActions);
+
+	const ref = useRef<HTMLDivElement>(null);
+	useEffect(() => {
+		if (scrollToItem === item.itemId) {
+			ref.current?.scrollIntoView({ behavior: 'smooth' });
+			setScrollToItem(null);
+		}
+	});
 
 	const ribbonColor = useItemColorRibbon([], wornItem ?? null);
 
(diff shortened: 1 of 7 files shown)


ClaudiaMia:
Hi!~

I tested it and it works very well. I really love it. This was badly missing!

One thing I noticed that I never noticed before: When you use the new button to jump to the room item view of the room inventory, the tab is weirdly cut off at the top. One can see that the text is not fully v-center, because 2 or 3 pixel lines are missing from the top, which is noticeable from the back button and the tab itself, being not fully round. I have no idea how this happened, and when you open the room inventory manually, it is not like this. But it is also a very minor thing.

About the design decision: I guess it may not be immediately obvious what this button does, but I feel it is simple, elegant and clear enough solution to me. A user will quickly learn what it is and does.

About technical points: For that we need Sekkmer or Jomshir to address these points, as this is not my speciality~

Thank you!~

purpleharmony:
Good catch! The vertical shift seems to be happening because of the autoscroll to the device in the room inventory list - it doesn't happen if I comment out the scrolling.
Right now I don't know how to fix it, I would have to look more into it. But maybe someone does?

purpleharmony:
I figured it out (well, kinda). I **think** what was happening was that the canvas was overflowing from its container, because canvas are `display: inline;` by default, and when the scroll event happened it scrolled the canvas in its container as well, hiding part of the tab.

pandora-client-web/src/components/wardrobe/views/wardrobeItemView.tsx
```diff
@@ -252,10 +252,18 @@ function InventoryItemViewList({ item, selected = false, setFocus, singleItemCon
...
 	const wornItem = useWardrobeTargetItem(target, item);
 	const extraActions = useObservable(extraItemActions);
 
+	const ref = useRef<HTMLDivElement>(null);
+	useEffect(() => {
+		if (scrollToItem === item.itemId) {
+			ref.current?.scrollIntoView({ behavior: 'smooth' });
+			setScrollToItem(null);
+		}
+	});
```
Sekkmer, line 265: there should be a dependency array parameter here, something like this:
`}, [scrollToItem, item.itemId, setScrollToItem])`

(eslint will show errors if something is missing)

pandora-client-web/src/components/wardrobe/wardrobeItems.tsx
```diff
@@ -98,8 +101,18 @@ export function useWardrobeItems(currentFocus: WardrobeFocus): {
...
+	const { target, targetSelector, assetList, heldItem, setHeldItem, focus, setScrollToItem } = useWardrobeContext();
+
+	const location = useLocation();
+	if (IsObject(location.state) && 'deviceId' in location.state) {
+		const focusItemId: ItemId = location.state.deviceId as ItemId;
+
+		focus.value = { container: [], itemId: focusItemId };
+		setScrollToItem(focusItemId);
+		location.state = {};
+	}
```
Sekkmer, line 113: I feel this should be inside a `useEffect`

Sekkmer (changes requested):

pandora-client-web/src/components/wardrobe/views/wardrobeItemView.tsx
purpleharmony: Sure. I remember removing the dependency array at some point because the linter was screaming at me, but with the current code it works.

pandora-client-web/src/components/wardrobe/wardrobeItems.tsx
purpleharmony: Done

purpleharmony:
Thanks for the review. I rebased on master.

Sekkmer (approved):
looks good for me

Issue #714, opened 2024-04-18
Clear kicked/banned players from device slots automatically


ClaudiaMia:
Currently, the "clear occupancy of the slot" is a way for admins to clean up after any sync issues, such as someone being kicked from the space. If they just kick and forget to clear the slot, then no one one else that is not admin can clear it and use it again.

I just tested it, and if the one kicked joins again, then even the admin can no longer clear the slot and they have to undeploy and deploy again. It is a bit of an inconvenient that this is a manual step that can easily be forgotten.

This issue is about investigating if admin action could have an automated clean-up step that clears the target from all slots in all devices of the space (by cycling through them).

Note: When Pandora's Space Service kicks some ghost out of a room device, it should also clear the slot.

ClaudiaMia:
https://github.com/Project-Pandora-Game/pandora/pull/722 contained an uncaught bug and had to be reverted, as described in https://github.com/Project-Pandora-Game/pandora/pull/736. So this issue is still open and the remaining issue in the solution needs a fix.

XenonRope:
I wrongly assumed that a character is removed from a space only when the character leaves the space or is kicked/banned. I reproduced the bug locally and I'm going to fix it.

Jomshir98:
Yes... All good about that bug.
I myself literally wrote the code in question, reviewed your PR, tested it locally before merging it; and still found out about the issue only after deploying it to production 😅.
We really are in need for tests for asset framework, which would catch problems like this that were once solved, but the existing solution is relatively fragile. (This exact problem of characters getting removed from room devices has happened few months back during development, but I forgot about it since; it also is the main reason for the current implementation)
I think the easiest way to go about this would be having an explicit cleanup method that would have to be called explicitly by Shard. But there is a problem on Shard that it currently doesn't differentiate between character moving from the space into another (possibly on a different shard) and between character being unloaded right before space unload (there is a valid invariant that space the character is in is always loaded when the character is, so space gets loaded before character, but unloaded only after character unload)

Sekkmer:
couldn't we just simply introduce a flag on unload that would disable this cleanup, so while the space unload or loads it doesn't clean up room devices only if the space is fully loaded

XenonRope:
Unfortunately, I won't have time to continue this issue so I will unassign myself.

ClaudiaMia:
> Unfortunately, I won't have time to continue this issue so I will unassign myself.

[...]

PR #720, opened 2024-04-20
[FEATURE] Autokick of ghost characters

Jomshir98:
Adds ability to configure automatic kick of offline characters.
The scan for offline characters happens regularly by Shard and checks if character has been offline for X minutes before requesting Directory to kick it. If the request fails for whatever reason, it will be re-attempted during the next cycle.
Also has an option to ignore Owners/Admins/VIPs of a space.

Additionally increased the randomness of initial position generation 5-fold.

### DIFF
--- a/pandora-client-web/.eslintrc.json
+++ b/pandora-client-web/.eslintrc.json
@@ -14,7 +14,12 @@
 		{
 			"files": ["*.ts", "*.tsx"],
 			"parserOptions": {
-				"project": ["./tsconfig.json", "./test/tsconfig.json", "./tsconfig.webpack.json"],
+				"project": [
+					"./tsconfig.json",
+					"./test/tsconfig.json",
+					"./test/tsconfig.setup.json",
+					"./tsconfig.webpack.json"
+				],
 				"sourceType": "module"
 			},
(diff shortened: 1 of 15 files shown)


Sekkmer (approved):

ClaudiaMia (changes requested):
I tested it a bit more and I have three improvement proposals to make this more usable:

1. It currently also kicks ghost in room devices. This is may be undesired sometimes and also leads to the need for another manual step due to the issue that I described in https://github.com/Project-Pandora-Game/pandora/issues/714

I propose to add another toggle "Also affect characters in the slots of room-level items" with a default off.

2. I think public and private spaces are a bit different, so another toggle "Also in effect when the space is private" with a default off.

3. Less of a priority, but nice to have: When you currently enter a value of "0", I would expect that it kicks instantly when the disconnect message is printed in the chat (ideally without a separate chat message to not be spammy). Currently, a value of "0" still means after 1 minute. This is very minor and likely too much effort to address right now, but I thought it may be a nice idea. (or we simply do not allow to enter any values below 1)

ClaudiaMia (approved):
LGTM!~

PR #722, opened 2024-04-21
[FIX] Clear occupied slots after character leaves space

### LINKED ISSUE #714: Clear kicked/banned players from device slots automatically

XenonRope:
Fixes #714 

Slots occupied by characters that left the space (e.g. kicked/banned) will be cleared automatically. The button "Clear occupancy of the slot" should never be visible to a player but I left this button just in case.

There is still one minor issue: If a character is restrained to Bondage Cross with "Module: Chains" set to "Tied front-facing" and the character is kicked, the "Module: Chains" is still set to "Tied front-facing". To use Bondage Cross again, the "Module: Chains" must be set to "Not attached" manually. I didn't change that because it seems to work as expected. I can create a new Bondage Cross and set "Module: Chains" to "Tied front-facing" even if no character is restrained (the slot is empty).


pandora-server-shard/src/spaces/space.ts
```diff
@@ -554,3 +556,23 @@ export abstract class Space extends ServerRoom<IShardClient> {
 function IsTargeted(message: IClientMessage): message is { type: 'chat' | 'ooc'; parts: IChatSegment[]; to: CharacterId; } {
 	return (message.type === 'chat' || message.type === 'ooc') && message.to !== undefined;
 }
+
+function ClearOccupiedSlotsByCharacter(globalState: AssetFrameworkGlobalState, characterId: CharacterId): AssetFrameworkGlobalState {
```
Sekkmer, line 560: I feel this code should not be shard specific, instead moved to `withCharacter` function

XenonRope: I agree. I moved this logic to `withCharacter` function and added two unit tests.

Sekkmer (approved):
Looks good for me.

pandora-common/src/assets/state/globalState.ts
```diff
@@ -172,7 +174,7 @@ export class AssetFrameworkGlobalState {
 		return new AssetFrameworkGlobalState(
 			this.assetManager,
 			newCharacters,
-			this.room,
+			newRoom,
```
Jomshir98, line 177: When a room state changes, all existing characters _must_ be updated using `character.updateRoomStateLink(newState, false)`, otherwise they will still have links to the old room state.
In this specific case, this would break things for multi-character room devices - when one of the characters leave, the device itself is updated, but the other character in it would still have internally link to the old state.
See `withRoomState` for the loop that needs to be added.

Jomshir98 (changes requested):
Hi! Looks good, except the one usecase that breaks multi-character devices if one of the characters leave. See my comment below for more details.

pandora-common/src/assets/state/globalState.ts
XenonRope: You're right, I didn't notice this. I created the method `_updateRoomStateLinkOnCharacters` and added its invocation here. Additionally, I added a unit test for exactly this case. To make tests more readable I added `stateUtils.ts` as well.

Jomshir98 (approved):
Looks good to me now! Thank you!

PR #723, opened 2024-04-21
[FIX] focus previous on lock slot

Sekkmer:
tested with and without `React.StrictMode`


ClaudiaMia (approved):
Since you said it is an issue that only appears in the production built, I cannot really test it, but since it is a bug fix of a known crash, it should go in sooner rather than later.

PR #728, opened 2024-04-23
[FEATURE] Sorting of room characters list

### LINKED ISSUE #716: Improve sorting of the member list in the room tab
Currently, the sorting of the member list is working in the way that your own character is at the top, and then the rest of the list comes.

The sorting could be improved in the following way:

1. Your own character
2. Characters from accounts on your contact list
3. Online characters
4. Offline characters

Optional bonus idea: Characters could get sorted by when they have been joining the room or reconnecting, so that someone joining/reconnecting last is always at the top within their respective group (except offline characters, as we likely have no true time stamp for sorting within different groups, so someone disconnecting would reset this ordering which is fine).

purpleharmony:
Fixes #716 

The proposed order is
- Player
- Online friends & online characters from player account
- Online non-friends
- Offline friends & offline characters from player account
- Offline non-friends

The ordering within groups is from last to join to first to join; because in JS sets preserve insertion order and sorting is stable.

Right now, the useMemo is kinda useless because `characters` changes everytime someone moves around in the space, causing the whole component as well as the useMemo to run again. I think it might be a good idea to separate the Character type in two in order to avoid re-rendering the whole component when the list hasn't changed.

### DIFF
--- a/pandora-client-web/src/ui/screens/room/roomControls.tsx
+++ b/pandora-client-web/src/ui/screens/room/roomControls.tsx
@@ -1,5 +1,5 @@
 import React, {
-	ReactElement, useEffect,
+	ReactElement, useEffect, useMemo,
 } from 'react';
 import { useLocation, useNavigate } from 'react-router-dom';
 import { Button } from '../../../components/common/button/button';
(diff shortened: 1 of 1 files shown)


Jomshir98 (approved):
Looks and works great!

Yes, the fact that the array changes every time makes the `useMemo` a bit useless... on the other hand if it didn't change the order wouldn't update when character goes online/offline, as accessing the online state using `character.data.isOnline` wouldn't trigger a render when it changes.
That said I believe it is good for now, even if the always-refreshing array gets changed, as just changing tab will cause the list to update even then. Also fixing this properly would require having a primitive for resolving observable for all elements in a dynamically sized array, which we don't have yet. So definitely "good enough" solution.
Also I have to say, I really like the way the code was written!

Sekkmer (approved):
looks good for me too

PR #734, opened 2024-04-25
[FEATURE] Clickable links in account/character profiles

### LINKED ISSUE #711: Making links in profiles clickable
Links in both profiles should be detected and made click-able so that we can give them the same treatment as links in the chat: A warning confirmation popup about opening an external page (reuse of the very same).

Livie53:
Closes #711.

Splits profile descriptions by line to replace links with `<UntrustedLink>` elements. Consequently, currently only detects links when they're at the start of a line.

### DIFF
--- a/pandora-client-web/src/components/profileScreens/profileDescription.tsx
+++ b/pandora-client-web/src/components/profileScreens/profileDescription.tsx
@@ -1,0 +1,37 @@
+import type { ReactElement } from 'react';
+import { SpaceInviteEmbed } from '../../ui/screens/spaceJoin/spaceJoin';
+import { UntrustedLink } from '../common/link/externalLink';
+import React from 'react';
+
+export function ProfileDescriptionLine({ line, index }: { line: string; index: number; }): ReactElement {
+	if (line.match(/^https?:\/\//)) {
+		const url = new URL(line);
+		switch (url.hostname) {
+			case 'project-pandora.com':
+			case 'www.project-pandora.com':
+				if (url.pathname.startsWith('/space/join/')) {
+					const invite = url.searchParams.get('invite') ?? undefined;
+					const spaceId = url.pathname.split('/').pop();
+					if (!spaceId)
+						break;
+
+					return <SpaceInviteEmbed key={ index } spaceId={ spaceId } invite={ invite } />;
+				}
+				break;
+		}
+		return (
+			<UntrustedLink key={ index } href={ line }>
+				{ line }
+			</UntrustedLink>
+		);
+	}
+	if (line.length === 0) {
+		// Keep single space for empty lines to preserve line-breaks.
+		return (
+			<span key={ index }> </span>
+		);
+	}
+	return (
+		<span key={ index }>{ line }</span>
+	);
+}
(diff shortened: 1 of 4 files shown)


pandora-client-web/src/components/profileScreens/profileDescription.tsx
```diff
@@ -0,0 +1,37 @@
...
+					return <SpaceInviteEmbed key={ index } spaceId={ spaceId } invite={ invite } />;
+				}
+				break;
+		}
+		return (
+			<UntrustedLink key={ index } href={ line }>
+				{ line }
+			</UntrustedLink>
+		);
+	}
```
Sekkmer, line 27: this is the same as in `RenderChatPart` we should only have a single function/component for this

Sekkmer (changes requested):

pandora-client-web/src/components/profileScreens/profileDescription.tsx
Livie53: Yeah, I'd meant to ask about that, but I ended up forgetting. Somewhat unsure on where the best place to have that function would be

Sekkmer: maybe next in spacejoin where the `SpaceInviteEmbed` is, also not quite sure

Livie53: Changed to move the URL handling logic into a common function, part of it is still replicated, but was as close as i figured it should be given the slight differences between `RenderChatPart` and the added `ProfileDescriptionLine`.

Sekkmer (approved):

purpleharmony:
The empty new file at `pandora-client-web/src/ui/components/common/linkConverter.ts` is probably not intentional

Livie53:
> The empty new file at `pandora-client-web/src/ui/components/common/linkConverter.ts` is probably not intentional

Yeah, that was a file left from some testing that I guess got `git add .`'d with the rest, removed.

Livie53:
Thought of a different, imo better way to handle links in description, requested re-review due to changes

pandora-client-web/src/components/profileScreens/profileDescription.tsx
```diff
@@ -0,0 +1,24 @@
...
+		</div>
+	);
+}
+
+function DescriptionSegment({ segment, index }: { segment: string; index: number; }): ReactElement {
+	if (segment.match(/^https?:\/\//)) {
+		const url = new URL(segment);
+		return (
+			<RenderedLink key={ index } index={ index } url={ url } />
+		);
```
Sekkmer, line 19: I feel it would be best to wrap this in a try/catch as `new URL('https://%')` would throw and cause the component to crash

`RenderChatPart` would also need it, it was just more noticeable when you edit a profile rather then pasting a link in chat

```ts
		try {
			const url = new URL(segment);
			return (
				<RenderedLink key={ index } index={ index } url={ url } />
			);
		} catch { /**/ }
```

Sekkmer (approved):

PR #737, opened 2024-04-25
[FEATURE] device module permission

### LINKED ISSUE #592: Add optional binding between room device modules and slots for permission check
Allow binding a room device module to a single slot optionally
When a module is bound to a slot and the slot is occupied changing the module should check for permissions on the character who occupies the slot

Same for lock modules: Those should respect the most restrictive value of any occupied slots' character permissions, in case the lock slot is ambiguous and not tied to a specific character slot.

So if a character has password locks blocked by permission then this lock slot module that is linked to the character slot should not allow this lock.

Note: Exclusive locks may also be a bit special with room devices. I guess they should prevent any character inside character slots linked to the lock module from opening the lock.

Sekkmer:
resolves #592

works for all module kinds in room devices

sadly `assetSpecidc: undefined` must be added to all modules as I couldn't think of a way to eliminate the property from the type, however since it is `undefined` it will not be part of the JSON output

linked modules will prompt for permission:
image


ClaudiaMia (changes requested):
Exclusive locks currently do not work despite the new binding. Does not matter who adds it. The person in the character slot that should be locked into the slot due to the lock module can simply unlock the exclusive lock.

Sekkmer:
> Exclusive locks currently do not work despite the new binding. Does not matter who adds it. The person in the character slot that should be locked into the slot due to the lock module can simply unlock the exclusive lock.

now that I think about it it may be due to the other bug that device modules do not effect the one inside

ClaudiaMia (approved):
Tested that exclusive locks now work correctly and also tested other things, such as deploying a device with a character inside that the admin has no permissions to interact with.

Found no issues.

Jomshir98 (approved):

PR #739, opened 2024-04-27
[REFACTOR] Improve code to use an exhaustive switch statement for the enum type with compile time checking

Angela-BC:
Resolve a left over note after #695


Sekkmer (approved):
looks good

Jomshir98 (approved):

Issue #704, opened 2024-04-15
Improve clarity of hints about lock input requirements


ClaudiaMia:
Our lock framework should stay as generic as possible, but we got two reports of the messages being not clear enough:
https://discord.com/channels/872284471611760720/872568378190086174/1229311366758924288

PR #746, opened 2024-05-01
[FIX] Improve clarity of hints about lock input requirements

Angela-BC:
See #704 

Uses the password format to determine part of the hint message.
Adds another message when the entered message length is less than min while min equals max.


pandora-client-web/src/components/wardrobe/modules/wardrobeModuleLockSlot.tsx
```diff
@@ -306,38 +306,49 @@ function PasswordInput({
 
 	const id = useId();
 
+	let inputCharacterType: string;
+	let replaceFunc: ((_: string) => string);
+
+	switch (password.format) {
+		case 'numeric':
+			inputCharacterType = 'digits';
+			replaceFunc = (v) => v.replace(/[^0-9]/g, '');
```
Sekkmer, line 315: unfortunately moving this out of `useCallback` is not ideal as lambda functions will return a new object on each render so the `useCallback` will have no effect as `replaceFunc` change every time

Sekkmer (changes requested):

pandora-client-web/src/components/wardrobe/modules/wardrobeModuleLockSlot.tsx
Angela-BC: I did go for useMemo to memoize the function and the string.
- I'm assuming using a tuple is not a problem?
- A little bit later in the code there is another switch on password.format and allow regex tests. I could also create the allow functions when I create the replace functions. This way the code is next to each other.

Sekkmer (approved):

pandora-client-web/src/components/wardrobe/modules/wardrobeModuleLockSlot.tsx
Sekkmer: yeps this works, tuples are ok

Issue #689, opened 2024-04-13
Improve long character slot selection lists


ClaudiaMia:
In specific room item placements, and depending on the canvas size and resolution, as well as portrait/landscape, the selection menu cannot fit the screen and will be cut off, meaning part of the list will not be usable. It is not draggable or scrollable.

![image](https://github.com/Project-Pandora-Game/pandora/assets/99583892/1d5ecdaf-a3c7-468b-9e7b-4754e9041649)

Jomshir98:
This has already been fixed in the above-mentioned PR.

PR #750, opened 2024-05-02
[FIX] Improve long character slot selection lists

Angela-BC:
See #689 

For now I did only adjust DeviceSlotsMenu. Are there other context menu's that needs adjustment?

This change has impact on **all** context menu's, since it sets the max-height in the context-menu css class. Is this going to be an issue?


Jomshir98 (changes requested):
Hi!
The changes in the context menu break the styling - the character buttons (given sufficiently long slot name, such as those of "Bondage Frame") are no longer stretched and no longer have spacing between them.
I experimented slightly and I suggest more generic fix that will add a scrollbar to _any_ contextmenu, if it gets too big:
- Keep the `max-height` style
- Wrap the content of the device context menu in `<Scrollable>` and `<Column>` so the tree looks like in the code below. This should apply to the following components:
  - `DeviceContextMenuCurrent`
  - `CharacterContextMenu`
  - `ContextMenuImpl`
- Together with that the added `<div>` should be removed from the `DeviceSlotsMenu`.

The code to produce scrollbar for the whole menu while keeping styling intact:
```tsx
<div className='context-menu' ... >
	<Scrollable color='lighter'>
		<Column>
			{ original content }
		</Column>
	</Scrollable>
</div>
```

Angela-BC:
That gets overlooked when the test characters have names like Angela, Dummy, Dummy two, Dummy Three and Dummy four. Looks Like I'd better get some test chars with a longer name.

I adjusted the code as suggested.

Jomshir98 (approved):
Looks good! Thank you~

ClaudiaMia (approved):
Tested and working great!~

PR #751, opened 2024-05-02
[FEATURE] item spawned by property

Sekkmer:
add `spwanedBy: CharacterId` to all items, property is server only and never sent to the client


Jomshir98 (approved):
LGTM!

PR #768, opened 2024-05-14
[FEATURE] Multiple texture resolutions support

Jomshir98:
**Depends on [pandora-assets#242](https://github.com/Project-Pandora-Game/pandora-assets/pull/242).**

Adds a support for using down-scaled textures.
At the moment resolutions 1 (no change), 1/2, and 1/4 can be selected. This applies to normal assets, room devices and room backgrounds; no matter which screen uses them (both room and wardrobe).

There is also an option for "auto", which tries to guess best resolution automatically (might need tweaking in the future). This is the default.
It works by looking at the true pixel size of the screen (so window size multiplied by DevicePixelRatio (a.k.a. factional scaling). It is tuned such that PC FHD (1080p) screen gets 1/2 resolution (it still looks rather good!), but QHD (1440p) gets full resolution.
For phones the logic works the exact same. To be specific there are thresholds for both width and height, combined with an 'OR'.
The automatic resolution is remembered and reused the next time. The screen is also observed for changes and the resolution might get automatically upgraded if the window is resized. It is never downgraded automatically.

### DIFF
--- a/pandora-client-web/src/assets/assetGraphics.ts
+++ b/pandora-client-web/src/assets/assetGraphics.ts
@@ -1,14 +1,15 @@
+import { Draft, Immutable, freeze, produce } from 'immer';
+import { cloneDeep, maxBy, minBy } from 'lodash';
 import { Assert, Asset, AssetGraphicsDefinition, AssetId, BoneName, CharacterSize, Item, LayerDefinition, LayerImageOverride, LayerImageSetting, LayerMirror, LayerPriority, PointDefinition } from 'pandora-common';
+import { useMemo } from 'react';
+import { AppearanceConditionEvaluator } from '../graphics/appearanceConditionEvaluator';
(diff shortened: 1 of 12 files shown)


Sekkmer (commented):
Code looks good

1. auto selection
as we support zoom it is highly likely users will not be fully zoomed out all the time, so does the "it still looks rather good!" covers zoom?

2. editor 
is this setting available from the editor and would it be possible to use default 1 so in editor you see everything as is by default, or maybe have a toggle at the source selection screen so you can pick it when starting

Jomshir98:
1. It covers zoom in the sense that it was tested with "The whole screen is focused on a one character" - so zoom to the level of whole character still being visible.
2. I will check that, it might need some changes.

Sekkmer (approved):

PR #770, opened 2024-05-15
[FEATURE] Add a command to manipulate the player's blush asset

SandrinePDR:
This adds the /blush chat command with the following options:
* '+' increase the blushing by one level
* '-' decrease the blushing by one level
* 'min' sets the blushing to the lowest available level
* 'max' maximizes the blushing

This should work, regardless of the chosen blush asset, in case we have more in the future.


pandora-client-web/src/ui/components/chat/commands.ts
```diff
@@ -368,4 +369,78 @@ export const COMMANDS: readonly IClientCommand<ICommandExecutionContextClient>[]
...
+		/*
+			This manipulates the blushing of a player. The main challenge here was the incredilble flexible way,
+			pandora handles assets. Currently we only have one blushing variant, but it may be that we will have
+			other variants with more or less levels than the current one. So this flexibility had to be dealt with
+			when the item was selected. Hopefully the way how 'blush' is extracted from the player can work
+			as a blueprint for others.
+		*/
+		description: 'Increase or decrease your blushing',
+		longDescription: '',
+		usage: 'blush + | - | min | max',
```
ClaudiaMia, line 384: Should be just `'+ | - | min | max'` as the "blush" is added automatically.

pandora-client-web/src/ui/components/chat/commands.ts
```diff
@@ -368,4 +369,78 @@ export const COMMANDS: readonly IClientCommand<ICommandExecutionContextClient>[]
...
+							newBlush = currentBlush > 0 ? variants[currentBlush - 1].id : variants[currentBlush].id;
+							break;
+						case 'min':
+							newBlush = variants[0].id;
+							break;
+						case 'max':
+							newBlush = variants[variants.length - 1].id;
+							break;
+						default: // This shouldn't happen, but hey...
+							return false;
```
Sekkmer, line 420: this should be an `AssertNever(options)` so the enum and the cases are kept in check by TypeScript

pandora-client-web/src/ui/components/chat/commands.ts
SandrinePDR: Changed.

pandora-client-web/src/ui/components/chat/commands.ts
SandrinePDR: Good point. I added that.

ClaudiaMia (approved):

Sekkmer (approved):

Issue #676, opened 2024-04-13
Investigate if sentry can be disabled from the hCaptcha component


ClaudiaMia:
Error after block by uBlock origin due to problematic component detected:
https://discord.com/channels/872284471611760720/872568378190086174/1227704403759468675

Jomshir98:
It would be also great to get hCaptcha library to lazy-load completely while at this.

PR #771, opened 2024-05-17
[CHANGE] lazy load hcaptcha

Sekkmer:
I also set `sentry={ false }` tried it locally and it works but not quite sure if this count as removing `Sentry` as the loader package doesn't use dynamic import so the source for sentry will still be present

ref: #676


Jomshir98 (approved):
Looks good to me! Thank you~

PR #772, opened 2024-05-18
[FEATURE] show lock password

### LINKED ISSUE #717: Show the password / combination value to the one who entered it for safety and convenience
Within https://github.com/Project-Pandora-Game/pandora/issues/491 we plan to store information about the one who created an item and on the one who last changed something about an item.

Potentially, there could be a custom property that stores who entered the value and then allows to do the following:
Showing the one who set a password / combination the value. 
Note: In the following I will only say password when I mean both.

This could be easy as we store the id of who locked a password. It just becomes tricky since we allow to lock a password with the previous set value, and showing it to the one locking it that way would leak the password.

This is fine for a combination lock where it is a bit silly that you cannot know the previous value (unless it is a digital lock where such a value could be secret), but for a password lock it could definitely be realistic to not know it.

So either 
1. we store the value of who inputted a password as a special data for just this kind of lock modules.
2. ~~we show the password also to the one who reused the last set value~~
3. ~~we remove the feature that you can lock a password again with the previous entered value~~

Sekkmer:
resolves #717

image

### DIFF
--- a/pandora-client-web/src/assets/appearanceValidation.ts
+++ b/pandora-client-web/src/assets/appearanceValidation.ts
@@ -152,10 +152,13 @@ export function RenderAppearanceActionProblem(assetManager: AssetManagerClient, 
 		const f = result.failure;
 		if (f.type === 'moduleActionFailure') {
 			if (f.reason.type === 'lockInteractionPrevented') {
-				if (f.reason.reason === 'wrongPassword') {
-					return `The password is incorrect.`;
+				switch (f.reason.reason) {
+					case 'wrongPassword':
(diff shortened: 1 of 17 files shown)


Jomshir98 (approved):
Looks and works great from the technical standpoint!

From UI/UX standpoint it does feel a bit clunky. The button doesn't really do what I would expect and some problem messages (namely "invalid password" message) don't get cleared when the button is used. The name "Show" also feel slightly confusing and the button is always shown (even if there never was a password set).
But overall I feel it is good enough for now.

PR #780, opened 2024-06-03
[FIX] display name change validates on account settings

### LINKED ISSUE #743: Bug: Display name format is not validated before attemt to save it
The account display name has a limited format, but client's UI doesn't validate it before attempting to set it.
As the message to change settings if one-off, there is no reaction about this failing as it gets silently rejected.
We need to add a warning in UI when the new name is not valid.

zoe-64:
Previously changing display name to above 32 or below 3 wouldn't generate a warning toast. Now it does. 

Fixed by using FormCreateStringValidator similar to registrationForm. 

fixes #743


Sekkmer (approved):
looks good

PR #781, opened 2024-06-06
[DEV] Bunch of development tooling improvements

Jomshir98:
For full list of changes see individual commits. The rough summary is:
- Allow easily using HTTPS by proxying all development traffic through webpack
- Use new VSCode feature to mark some files as readonly (mainly inside any `dist` and `node_modules`)
- Add `data-action*` attributes to wardrobe buttons so their actions can be easily read while running at least `USER_DEBUG` client
- Speed up client compilation using fork-ts-checker-webpack-plugin
- Make E2E tests work on Windows (coverage is not working, but everything else works), while not breaking what already works (this was harder than it sounds)


pandora-client-web/.env.template
```diff
@@ -1,6 +1,5 @@
 # Server connection config
-DIRECTORY_ADDRESS="http://127.0.0.1:25560"
-#DIRECTORY_ADDRESS="https://directory.project-pandora.com"
```
Sekkmer, line 3: is there any reason not leave this here commented out?

Sekkmer (approved):
looks good

pandora-client-web/.env.template
Jomshir98: The reason is that as we made the production server policies more strict, it will no longer work - server will reject the connection from non-production client.

PR #784, opened 2024-06-19
[FEATURE] Add text that is displayed upon entering to a space

SandrinePDR:
Initial version. Entry text can be defined and is displayed, when a character enters a space


pandora-client-web/src/ui/screens/spaceConfiguration/spaceConfiguration.tsx
```diff
@@ -60,11 +61,13 @@ import './spaceConfiguration.scss';
 
 const IsValidName = ZodMatcher(SpaceBaseInfoSchema.shape.name);
 const IsValidDescription = ZodMatcher(SpaceBaseInfoSchema.shape.description);
+const IsValidFluffText = ZodMatcher(SpaceBaseInfoSchema.shape.fluffText);
 
 function DefaultConfig(): SpaceDirectoryConfig {
 	return {
 		name: '',
 		description: '',
+		fluffText: '',
```
Sekkmer, line 70: why is it called `fluffText` I wouldn't be able to guess what it is used for unless I also find the label for the user input, maybe it could be renamed to something more intuitive like:

`welcomeText`

SandrinePDR: Oh that's just me. "Fluff text" is text used on board games to fluff out game components without addin new mechanics or rules. But I understand that it is confusing and will change it.

Sekkmer (approved):
looks good for me^^

(as per review rules please wait for a second review or until 3 days has passed since the creation of this PR which should be tomorrow before merging)

Jomshir98 (changes requested):
Looks and works mostly good. There are, however some problems; noted all of them below.

pandora-client-web/src/ui/screens/spaceConfiguration/spaceConfiguration.tsx
```diff
@@ -208,10 +211,20 @@ export function SpaceConfiguration({ creation = false }: { creation?: boolean; }
 						value={ currentConfig.description }
 						onChange={ (event) => setModifiedData({ description: event.target.value }) }
 						readOnly={ !canEdit }
-						rows={ 16 }
+						rows={ 8 }
```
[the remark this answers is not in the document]
SandrinePDR: I changed the size of the description from 16 to 8 to maintain the size of the dialog. Should we rather decrease the other size as well or risk a scrollbar for the space creation dialog?
Also: wouldn't it be a good idea to use a constant for this value?

Jomshir98 (approved):
Looks good to me!

PR #786, opened 2024-06-21
[ADD] Add a scrollbar to the list of contacts

SandrinePDR:
If the list of contacts grows too long, one couldn't scroll the list. This issue is fixed with this PR.

### DIFF
--- a/pandora-client-web/src/components/accountContacts/accountContacts.tsx
+++ b/pandora-client-web/src/components/accountContacts/accountContacts.tsx
@@ -12,6 +12,7 @@ import { DivContainer, Row } from '../common/container/container';
 import { AccountContactChangeHandleResult, useFriendStatus, useAccountContacts } from './accountContactContext';
 import { useConfirmDialog } from '../dialog/dialog';
 import { useKeyDownEvent } from '../../common/useKeyDownEvent';
+import { Scrollable } from '../common/scrollbar/scrollbar';
 import './accountContacts.scss';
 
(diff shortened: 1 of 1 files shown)


pandora-client-web/src/components/accountContacts/accountContacts.tsx
```diff
@@ -178,31 +178,33 @@ function ShowFriends() {
...
-					<th>Actions</th>
-				</tr>
-			</thead>
-			<tbody>
-				{ friendsWithStatus.map((friend) => (
-					<FriendRow key={ friend.id } { ...friend } />
-				)) }
-			</tbody>
-		</table>
+		<div style={ { overflow: 'auto', height: '100%' } }>
```
Sekkmer, line 181: how about using `<Scrollable direction='vertical'>` instead?

(may still need height, not sure)

Sekkmer (commented):
good catch :)

thought we had an issue but I could not find any, I guess it was only mentioned on discord and we forgot to make an issue here

pandora-client-web/src/components/accountContacts/accountContacts.tsx
SandrinePDR: Good idea. I didn't know that one existed...
Why does the Scrollbar on directMessages.tsx#31 not work for the contacts?

Jomshir98 (approved):
Looks good! Thank you.

PR #792, opened 2024-07-02
[FIX] Fix Direct Messages key unlocking being case sensitive

Jomshir98:
## References

_None_

## About The Pull Request

This PR reworks how client handles Direct Messages in aim to make it much more resilient to failing to unlock the cryptographic key. The primary reason to start that work is a bug with current key format that treats username case-sensitively, but completely breaks down without reasonable user errors if user logs in with differently cased username.

In doing of this I made following changes:
- DMs are stored in encrypted form and (for the most part) not decrypted eagerly. The decryption is cached, however.
- Both `DirectMessageManager` and `DirectMessageChannel` were rewritten from scratch, copying most of the old code into them while tweaking it.
- Both now allow having and manipulating message data without loaded encryption; the `onMessage` handlers are now synchronous.
- Moved the notification suppression from using `mounted` counter to a `useNotificationSupress` in the component.
- Added a few GUIs for showing error states and attempting to unlock a key if that didn't succeed automatically with data from login.
- The key gets re-saved with a new password hash format after successful unlock.

## Changelog

```md
Fixes:
- Fixed an issue with Direct Messages that caused them to break if you logged in using username with different CaPtaliZation.
  - Reworked the way Direct Messages handling works on the background to achieve hopefully smooth migration of existing data. If you have any problems with DMs after this update, please contact us.
  - You might get a prompt to unlock your DM cryptographic key the first time you enter Direct Messages screen, in order to migrate to a new key format.
```

## Checklist


pandora-server-directory/src/account/accountSecure.ts
```diff
@@ -253,8 +253,8 @@ export default class AccountSecure {
 		return this.#secure.cryptoKey?.publicKey;
 	}
 
-	public async setInitialCryptoKey(key: IAccountCryptoKey): Promise<'ok' | 'invalid' | 'keyAlreadySet'> {
-		if (this.#secure.cryptoKey != null)
+	public async setCryptoKey(key: IAccountCryptoKey, allowReset: boolean = false): Promise<'ok' | 'invalid' | 'keyAlreadySet'> {
+		if (this.#secure.cryptoKey != null && !allowReset)
```
Sekkmer, line 257: hmm, maybe it would worth checking that on reset we have the same public key if not all previous DMs would go invalid

Sekkmer (approved):
looks good, tested it locally and it worked as expected

PR #851, opened 2024-10-24
[FEATURE] Add random periodic blinking as a condition

### LINKED ISSUE #143: Add random blinking to the characters
To make the characters look more lively, a random blink should be added to them.
The average human blinks 10 to 15 times per minute, so a random period of time between 4 to 6 seconds after a blink would look natural.

Livie53:
## References

resolves #143 

## About The Pull Request

Adds a blinking condition intended for eye assets, randomised between 4-6 seconds after every blink cycle. Does nothing without asset overrides to use the added condition. Blinking should only apply for room views, currently excludes wardrobe, editor, etc. views.

## Changelog

Authored by: Livie53

```md
Platform changes:
- Added periodic blinking for characters.
```

## Checklist


pandora-client-web/src/graphics/graphicsCharacter.tsx
```diff
@@ -99,6 +107,35 @@ function GraphicsCharacterWithManagerImpl({
...
+				newBlinkSeed();
+			}, BLINK_LENGTH);
+		}, blinkInterval);
+
+		return () => {
+			clearInterval(intervalId);
+			clearTimeout(timeoutId);
+		};
+	}, [setCharacterBlinking, blinkRandom, newBlinkSeed, useBlinking]);
+
```
Jomshir98, line 138: This works, but there are a few small problems that can be improved upon:

1: `setInterval` can be replaced by `setTimeout` too. When the interval triggers, then you start a timeout for resetting blink (which is through the current values expected to trigger sooner than the outer interval). This eyes-open timeout will then call `newBlinkSeed`, which will (through cascade) regenerate `blinkRandom`, which then triggers update of this `useEffect`, clearing the interval.
In short the interval never actually triggers more than once (if everything goes well).

2: Completely regenerating the random value generator each blink is technically equivalent to just generating a fully random number each blink.
Instead I would suggest generating pseudo-random generator only once, for example using:
```ts
const blinkRandom = useMemo(() => new PseudoRandom(nanoid()), []);
```
And then having the `useEffect` do another `setTimeout` whenever the eyes get opened (getting rid of the `useReducer` altogether). That way the `useEffect`'s dependencies never change, and it isn't cleaned up and re-run each blink.
Technically this could be achieved through, for example, creating a function inside the `useEffect` and then triggering it whenever it is appropriate. Rough outline I'm thinking about:
```ts
useEffect(() => {
    let timeoutId: number | undefined;
    const run = () => {
        const blinkInterval = ...
        timeoutId = setTimeout(() => {
            // Blink here
            timeoutId = setTimeout(() => {
                // Un-blink here
                run(); // loop
            }, ...);
        }, blinkInterval);
    };
    run(); // Trigger first loop
    // Cleanup handler
    return () => {
        // Cancel `timeoutId` if set
    };
}, [...]);
```

pandora-client-web/src/graphics/graphicsCharacter.tsx
```diff
@@ -76,6 +79,10 @@ export const CHARACTER_PIVOT_POSITION: Readonly<PointLike> = {
 	y: 1290, // The position where heels seemingly touch the floor
 };
 
+const MIN_BLINK_INTERVAL = 4_000; // ms
+const MAX_BLINK_INTERVAL = 6_000; // ms
+const BLINK_LENGTH = 250; // ms
```
Jomshir98, line 84: Idea: Should the length of the blink be randomized too? Humans blink for 0.1 to 0.4 apparently, and I think randomizing this too might make it look more natural.

Jomshir98 (changes requested):
Looks good and works! Nicely done!
I left several notes that could be improved upon before this is merged.

As for the part about adding documentation, here is what I would find sufficient:
- Add a `/** <explanation> */` comment for `AppearanceConditionEvaluator`'s `blinking` property, just saying that it is whether the character is in the middle of a blink
- Same for `useAppearanceConditionEvaluator` (sorry about there being no comment at all currently, if you type `/**` and select VSCode's suggestion, it should auto-generate a comment template for you)
- Comment the meaning and default value of `useBlinking` property in `GraphicsCharacterProps`
- Comment the meaning and default vault of `characterBlinking` property in `GraphicsLayerProps`
I don't think any more documentation is needed.
If you want to be extra thorough you could add more wordy explanation near `AtomicConditionBlinkingSchema`, but that is optional in my eyes.

pandora-client-web/src/editor/parsing.ts
```diff
@@ -76,6 +89,15 @@ function ParseAtomicCondition(input: string, validBones: string[]): AtomicCondit
...
 	}
+	if (input.startsWith('blinking_')) {
+		const blinking = /^blinking_(true|false)$/i.exec(input);
+		if (!blinking) {
+			throw new Error(`Failed to parse blinking condition '${input}'`);
+		}
+		return {
+			blinking: ParseBoolean(blinking[1]),
+		};
+	}
```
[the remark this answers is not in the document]
Livie53: The `_false|true` does feel a little awkward, from an editor user standpoint, the reasoning was to follow the general pattern of other potential inputs, as well as the potentially messy check for `!?blinking`. Still, I do believe it makes more sense to change it as suggested, and have changed it accordingly.

pandora-client-web/src/graphics/room/roomCharacter.tsx
```diff
@@ -422,6 +422,7 @@ const RoomCharacterDisplay = React.forwardRef(function RoomCharacterDisplay({
 					scale={ { x: scaleX, y: 1 } }
 					pivot={ pivot }
 					angle={ rotationAngle }
+					useBlinking={ true }
```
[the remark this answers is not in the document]
Livie53: A consequence of unfamiliarity with React's properties. Fixed.

pandora-common/src/assets/graphics/conditions.ts
```diff
@@ -74,6 +76,9 @@ export const AtomicConditionLegsSchema = z.object({
 export const AtomicConditionViewSchema = z.object({
 	view: CharacterViewSchema,
 });
+export const AtomicConditionBlinkingSchema = z.object({
+	blinking: BlinkingSchema,
+});
```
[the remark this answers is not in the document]
Livie53: I wasn't entirely sure whether it should be a simple boolean, or to follow the separation as with other conditions, so I had decided on the latter to be safe. As it's likely to be unnecessary then, it has been changed for simplicity.

pandora-client-web/src/graphics/graphicsCharacter.tsx
Livie53: > `setInterval` can be replaced by `setTimeout` too
I had noticed this while looking through the commits, but evidently forgot to actually change it... Fixed now.

> 2
A very fair point I had not considered. Initially, it had been achieved through a looped function call, before moving it inside the effect. Had forgotten about inline functions. Logic has been changed accordingly, and assuming I understood correctly, the PRNG should now only be created once.

pandora-client-web/src/graphics/graphicsCharacter.tsx
Livie53: I had wondered earlier if it should be or not, but since it added not entirely necessary complexity outside of the given spec, I decided against it for now. Since it's a relatively easy change though, it is now randomized as well.

Livie53:
Thanks for the review! I tried adding documentation to at least the affected properties/changes, may still need cleanup though. Ready for re-review, no further changes planned at least as of writing.

Jomshir98 (approved):
Looks and works great!

PR #855, opened 2024-11-09
[FIX] Move graphics draw props to callbacks 

Livie53:
## References

_None_

## Changelog

Authored by: Clare & Claudia

```md
Platform changes:
- Added ability to show current passwords for locks for the person who set it
- Reworked the way login tokens work in Pandora:
  - Logging out in one tab now logs you out of the same account in all other tabs of the same browser
  - Pandora is now supporting login tokens over multiple devices

Fixes:
- Fixed the target name of a kick message being the one who initiated it
- Fixed an issue that could lead to an inaccurate display of online status in the contacts list

Technical changes:
- Improved tooling related to development and testing; you can find more details [in the original PR](<https://github.com/Project-Pandora-Game/pandora/pull/781>)
```

## Checklist

- [ ] The change has been tested locally
- [ ] Added documentation to the new code and updated existing documentation where needed
- [ ] I understand this patch is submitted under the [Pandora's Contributor Agreement](https://github.com/Project-Pandora-Game/pandora/blob/master/contributor-licence-agreement.md)

### DIFF
--- a/pandora-client-web/src/graphics/room/roomCharacter.tsx
+++ b/pandora-client-web/src/graphics/room/roomCharacter.tsx
@@ -381,6 +381,28 @@ const RoomCharacterDisplay = React.forwardRef(function RoomCharacterDisplay({
 			AssertNever(interfaceChatroomCharacterNameFontSize);
 	}
 
+	const graphicsDraw = useCallback((g: PIXI.GraphicsContext) => {
+		g
+			// Pivot point (with extra Y offset)
+			.circle(pivot.x, pivot.y, 5)
+			.fill(0xffaa00)
+			.stroke({ color: 0x000000, width: 1 })
+			// Mask area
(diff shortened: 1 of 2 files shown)


Jomshir98:
This certainly is an efficiency improvement, but I would argue it isn't a memory leak and that it definitely doesn't affect normal users.
1) All the modified code is only used if `debugConfig?.deviceDebugOverlay` is `true`, which isn't the case for regular users
2) The effect of the functions being recreated and re-assigned each render of the parent component only means that the child needs to be updated and re-drawn more often. It still shouldn't affect the child before the change getting properly cleaned up - so no memory leak should occur.

Either way it is an improvement and I will look at it in more detail tomorrow, when I'm less sleepy.

Jomshir98 (approved):
Although this doesn't fix the memory leak, it still is an improvement. I updated the changelog to reflect the actual effect of this change.

PR #862, opened 2024-12-02
[FEATURE] Graphics transitions

Jomshir98:
## References

_None_

## About The Pull Request

This PR adds smooth transitions for when something changes (character moves or changes pose).
Currently this is enabled only for characters (not for room devices or characters _in_ room devices) and only for the main room view (not in wardrobe).
Also implements setting to disable this, which defaults to "auto" = enabled unless system-wide "prefer-reduced-motion" is in effect.

## Changelog

```md
Platform changes:
- Added transition animations to characters in the room:
  - Moving or changing character's pose will now smoothly transition to the final pose instead of jumping instantly
  - The transitions respect item limits, so if your movement is limited you might still jump instead of transitioning between the poses
  - If you are in movement or posing mode, the transitions are very short to still let you precisely position your character
  - Transitions are currently enabled only for stand-alone characters - no transitions happen for room devices or characters inside room devices
  - A setting is available to disable this feature; it is also disabled by default if you have "prefer reduced motion" accessibility setting enabled
- Unified update throttling across Pandora: All of movement, posing, and item coloring now allow for up to 8 updates per second
```

## Checklist

### DIFF
--- a/pandora-client-web/src/components/settings/graphicsSettings.tsx
+++ b/pandora-client-web/src/components/settings/graphicsSettings.tsx
@@ -1,0 +1,207 @@
+import { FormatBytes } from 'pandora-common';
+import React, { ReactElement, useCallback, useEffect, useMemo, useState } from 'react';
+import { z } from 'zod';
+import { GraphicsManagerInstance, type IGraphicsLoaderStats } from '../../assets/graphicsManager';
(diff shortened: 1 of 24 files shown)


ClaudiaMia (approved):
Yay!~

PR #863, opened 2024-12-03
[FIX] Bunch of fixes regarding DMs and encryption keys changing

Jomshir98:
## References

_None_

## About The Pull Request

Fixes bunch of problems regarding direct messages and encryption keys changing. See individual commits or changelog for more details.

## Changelog

```md
Platform changes:
- Added warning to the password reset form, that resetting your password will clear all Direct Message history for both you and your conversation partners

Fixes:
- Several fixes for Direct Messages:
  - Fixed history of direct messages not loading correctly if you got a direct message before ever opening direct messaging window during the current session
  - Fixed properly showing when some messages in the conversation history are encrypted using old key (from before password reset), instead of showing a generic error
  - Fixed server "forgetting" all but the last direct message for conversations that were started before either of you reset your password but continued after password reset
  - Fixed trying to interact with accounts without valid encryption key appearing as if they didn't exist at all, instead of more sensible error message
  - Added several checks against trying to send direct messages to someone who is in the middle of resetting their password
  - Fixed rare case where you could miss incoming direct message if you briefly disconnected while looking at the direct messages screen and your conversation partner sent the message at that exact moment
```

## Checklist

### DIFF
--- a/pandora-client-web/src/components/common/form/form.scss
+++ b/pandora-client-web/src/components/common/form/form.scss
@@ -51,4 +51,8 @@
 		font-size: 0.95rem;
 		margin-top: 1em;
 	}
+
+	.warning-box {
+		margin: 0 0 1.5em 0;
+	}
 }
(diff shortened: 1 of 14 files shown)


ClaudiaMia (approved):
Great work!

PR #870, opened 2024-12-17
[ADD] More pose limit options for assets, resize viewport on overlay, and a few fixes

Jomshir98:
## References

_None_

## About The Pull Request

See changelog for individual changes.

## Changelog

```md
Platform changes:
- Pandora now resizes when viewport changes due to overlay (e.g. virtual keyboard). This means, that Pandora should "dodge" the virtual keyboard's opening and should no longer be overlaid by it

Fixes:
- Fixed attribute icons being black in some parts of the wardrobe, instead of being white (most notably, if you use attribute previews for items)
- Fixed local development environment sometimes breaking connection to the local development server, if running in the default "proxy" configuration

Technical changes:
- Assets now have more ways to define pose limits, allowing for more compact forms in some cases:
  - You can specify multiple independent pose limits now (useful if you want combination of several `options`)
  - You can define multiple string values (such as fingers state) at the same time (instead of needing to use `options`)
- Normal assets can now define custom pose presets as well (previously only room devices could)
```

## Checklist


ClaudiaMia (approved):
LGTM!

PR #878, opened 2025-01-06
[FEATURE] Action attemps and related changes

Jomshir98:
## References

_None_

## About The Pull Request

This PR introduces a new mechanic "Attempted action", that allows first starting an action and later finishing it.
A character can be attempting only one action at a time.

## Changelog

```md
Platform changes:
- Added a new setting "Bound Usage" to all wearable and room-level items that defines if this item can be used with restrained/blocked hands or not. This will allow a wider variety of roleplaying more conveniently, such as safe self-bondage, switch plays, and escape attempts.
  - Every item comes with a default that can be changed at item creation time or by anyone with permission afterwards.
  - There is a new interface setting that allows to override the default setting for all newly created items to the desired one.
  - When a character with blocked hands tries to use an action that is only possible due to the bound usage setting on an item, they will enter an "attempted usage" state (more on this lower).
  - Added a new entry with more details about this feature to the "Item" tab of the Wiki.
  - Spawning and deleting items, and changing this setting on existing items always requires free hands.
- Added a new "attempted usage" mechanic.
  - This can be seen in the wardrobe and chat by everyone in the room.
  - This state lasts indefinitely until the bound user decides to stop the attempt, finishes the action, or someone else interrupts it.
  - In the future we are planning more uses of this mechanic than "Using item while your hands are bound".
- Several improvements to the permission prompt dialog:
  - Much better description of what action the other character is requesting permission for.
  - Prevented prompt spam by the other character trying to perform the action many times - now only the last attempt is shown.
  - Added a small delay to confirming the action to avoid accidentally clicking on the confirm button when it appears/changes.
- You can now customize name and description of items actively worn by other characters. This comes with a new permission to allow this.
- Improved the behavior of custom item name and description editing, especially around cases where permission prompt is needed.
- Improved the behavior of posing UI and body resizing UI on other characters if you do not have a permission to do so.

Fixes:
- Fixed rare "failed assertion" error when trying to export data from Pandora.

Technical changes:
- Restrictions manager now reports all the restrictions instead of only the first one.
- Removed old `setView` action in favor of always using more generic `pose` action.
- There is now a clear distinction between body parts and wearable items (before they were both wearable items with different properties). This comes with a tiny change to asset creation, where you now use `DefineBodypart` instead of `DefineAsset` for body parts.
- Split up the appearance actions and moved them under action logic to clean up the, by now, huge file (sorry, @SandrinePDR, but it really was needed; I will fix the merge conflicts on your PR myself)
```

## Checklist


ClaudiaMia (approved):
LGTM and works great!~

PR #891, opened 2025-01-28
[ADD] Allow for positioning arms above or below hair

Jomshir98:
## References

_None_

## About The Pull Request

Allows for positioning arms above front hair or below back hair. These two options are marked as experimental, as they break with some hairs - this is likely  to be (mostly) resolved by arm split at elbow.

## Changelog

```md
Platform changes:
- Added ability to freely position your hands above or below front and back hair.
  - Note, that some combinations are marked as "experimental" and can lead to visual glitches, particularly if using long hair. While we are aware of this problem, there is no easy fix for it. Despite that we felt it is better to give the option to do so anyway, as there are cases where this works well.
- Updated several appearance validation messages to better convey that the action _would_ create an invalid state instead of referring to the current state.
- Added a note to the Wiki, that creating and deleting items always requires free hands.
- Added a tip about using character menu through the "Room" tab.
```

## Checklist

### DIFF
--- a/pandora-client-web/src/assets/appearanceValidation.ts
+++ b/pandora-client-web/src/assets/appearanceValidation.ts
@@ -128,13 +128,13 @@ export function RenderAppearanceActionProblem(assetManager: AssetManagerClient, 
 			}
 			case 'invalidState': {
 				if (e.asset) {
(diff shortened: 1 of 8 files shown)


ClaudiaMia (approved):
LGTM!

PR #895, opened 2025-02-08
[FIX] Update corepack npm registry keys

Jomshir98:
## References

_None_

## About The Pull Request

This fixes recent screwup of corepack shipping pinned keys, while npm registry rotated them.
I attempted to mitigate this into the future by just updating them every time from the npm endpoint.
This -technically- decreases security as they are no longer pinned, but we instead pin the hash of pnpm itself in our `packageManager` field.

## Changelog

_None_

## Checklist


ClaudiaMia (approved):
I am all for the usage of keys!

PR #905, opened 2025-02-20
[REFACTOR] Separate lock logic from lock items

Jomshir98:
## References

_None_

## About The Pull Request

Please see individual commits.

## Changelog

Authored by: Clare & Claudia

```md
Platform changes:
- Added a new information message about assets being updated

Fixes:
- Locking and unlocking locks now properly checks for having permission to use the specific lock type (previously only adding and removing the lock checked this and anyone could lock any lock if they could access the item it was on)
- Password locks no longer require inputting arbitrary password while in a safemode in order to unlock them
- "Invalid password" prompt from last unlocking attempt now properly disappears if you put in different password
- Clicking the up/down buttons on numeric inputs without having it selected already will now properly focus it

Technical changes:
- Refactored how locks work internally to be able to reuse their logic for an upcoming feature. Please let us know if you spot any new weirdness regarding locks!
- Optimized code related to loading assets while changing spaces, if the source and destination space use same asset definitions (which is always the case on the public version)
```

## Checklist

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -328,6 +328,7 @@ export default tsEslint.config(
 		],
 		extends: [
 			eslintReactPlugin.configs.flat.recommended,
+			eslintReactPlugin.configs.flat['jsx-runtime'],
 		],
 		plugins: {
 			'react': eslintReactPlugin,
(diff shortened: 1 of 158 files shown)


ClaudiaMia (approved):
LGTM!~

PR #907, opened 2025-02-20
[FEATURE] Character modifiers

### LINKED ISSUE #875: Adding a system for managing character specific rules that can customize how Pandora works
Rules are configurable modifiers that alter how Pandora's features work for the character a specific rule is in effect for.

Every character will have the same pool of rule types that allow to modify a specific feature of Pandora or may even introduce a new one.
For instance, a rule could allow to modify the initial delay before struggling out of an item is possible or allow to add a cooldown before struggling becomes possible again, after being interrupted by someone else. Or a rule could overlay a blur effect of configurable strength over the whole room canvas unless the character with this rule wears glasses.
Or a rule "prevent entering spaces with specific features" that is highly configurable so you can just forbid entering spaces in violation.

The same rule type template can be used to create multiple instances of rules of this rule type.
As this could lead to conflicts between rules, added character rules will be in an ordered list, where rules sorted "first" will have higher priority.

An added rule can be set to "on" or "off". 

A rule that is "on" is not necessarily always in effect, but only if certain conditions trigger it. These activation conditions shall use disjunctive normal form (DNF) to be highly flexible. The UI should make it easy and self-explanatory for the user to create custom activation conditions. Examples from good UIs that use DNF exist.
Essentially, "You have list of conditions. Each condition is of some type and might have some settings. You can have more conditions of the same type, as you always add them manually. Order matters. And each condition defines whether it is "AND" or "OR" with the previous ones, but it is in Disjunctive Normal Form. Meaning a chain of "AND" conditions has to be true together, but "OR" breaks the AND chain, starting a new AND chain. Any one AND chain is enough."

An extending idea: "Conditions presets"~
With two parts to the idea:
- You can export/import conditions
- Any rule type in Pandora could come with one or more rule-type-specific default condition presets that you can choose conveniently
  - Example: "Blur vision" having preset "while not wearing glasses" (not | has item with attribute | Glasses)
  
How rules interaction with the permission system:
- Each rule type would come with its own permission. All rules will by default be set to "Ask" or "Forbidden" based on how intense the rule is seen.
- You can give the same levels item limits use (favorite, normal, maybe, no, disabled) to each rule as well, but most are only visual - only "disabled" would have an effect, meaning that it simply cannot be added, even by yourself.
- There is a global permission "allow adding rules", that might even be blocked by default (despite many rules being "ask" by default)
- Likely also a good idea: Having a global switch "enable rules subsystem" that makes rules disabled altogether by default

And then for each rule instance:
- Order of rule instances matters (higher has priority)
- New rule instance is always created at the end
- To move a rule you need to be able to access all the rule instances between original and new position
- Each rule instance can be switched on/off
- Being in safemode allows you to deactivate rules that you would normally not be allowed to deactivate. There could be a rule that forbids you from deleting rules, so deactivating this one will allow you to delete rules.
- Each rule instance has editable conditions
  - We can consider timer conditions (maybe even instead of a "normal" timer like we had in BCX?)

More feature ideas:
"Locked rules"
  - You can imagine it as being able to add a lock on the rule - locking specific rule down no matter its permission
  - Will have some settings on its own (like who can get through the protection, if it has timer, ...)
  - Will require a separate global permission
  
"Suppressable rules"
Being able to declare some rule types as "suppress-able" from the start. This would mean that the rules have some external factor that just prevents them from working, by design. Example could be that specific spaces are able to disable a category of rules working inside. Or OOC-affecting rules being suppressed by not having a character selected.
That way we are very clear that some really intense rules can be "bypassed" in a way, and that fact being clearly communicated and explained.
The rule being suppressed does not unlock it, or change its config in any way.
Entering safemode or timeout mode will suppress every active single rule on the character.

Jomshir98:
## References

resolves #875

## About The Pull Request

This PR implements whole new framework for "Character Modifiers", which are mostly equivalent to BCX rules. We picked this name as we feel it better represents their broader purpose.
The main differences compared to BCX:
- Each modifier type can have multiple instances
- Advanced activation condition framework, heavily inspired by Factorio
- Each modifier type has its own permission, allowing for finer granularity than "normal/limited/blocked"
- Each modifier instance can be "locked" separately, allowing for finer control than permission-based one

## Changelog

```md
Platform changes:
- Added a new major feature to the wardrobe: "Character modifiers". While having the according permission, these can be added to a character and alter how Pandora's features work for this character while those modifiers are in effect.
  - Character modifiers represent a toolbox that can give more options to enhance your roleplay with effects and features beyond the standard. For example, a modifier could blur the room view while not wearing glasses.
  - There are three new permissions about viewing, editing, and locking character modifiers in general. The latter two are set to "blocked" per default, so this whole feature is opt-in.
  - The release comes with a first batch of 26 character modifier types and additionally 6 preconfigured character modifier templates for inspiration on what is possible with this feature.
  - Each modifier type comes with its own permission on top of the three general character modifier permissions.
  - A modifier type can be added more than once.
  - Enabled modifiers are in effect based on activation conditions, which can be customized based on 7 condition templates for now.
  - Added character modifiers can be exported to store them locally or to share them
  - Exported modifiers can be fully imported again or partially imported to reuse just the activation conditions on existing modifiers
  - There is a new entry on this feature in the wiki (characters-tab) that explains everything in more detail.
- Improved styling of the permissions list in settings for better usability
- Reworked the permission character selection overlay for better usability
- Changed the maximum allowed length of space, item, and outfit names from 40 to 48 characters
- Identical action messages by the server are now no longer repeated in direct succession but are denoted with a "+1" at the end to not clutter the chat.

Fixes:
- Fixed the long-hold issue on mobile devices when pressing on buttons with images
- Fixed the issue that draggable dialogs were able to appear offscreen (partially)
```

## TODO

This PR is work in progress. Currently remaining things:
- [x] Modifier locking
  - [x] Make locks work
  - [x] Lock bypass list
- [x] Modifier reordering checks
- [x] Add viewing help from modifier details
- [x] Document it in wiki
- [x] Write a changelog
- [x] More conditions
  - [x] Is wearing item with specified attribute
  - [x] Is wearing item named specific way
  - [x] Is affected by an effect (of strength at least ...)
- [x] More modifier types
  - [x] Speech modifiers
  - [x] Selective hearing impairment modifiers
  - [x] Blur effect
- [x] Delete `effect_blindfolds_blind_fully` once it can be replaced by conditions
- [x] Importing modifiers
- [x] Show which modifier types one has access to at a glance
- [x] Import of conditions (via exported of modifier template and partial importing)

## Checklist

- [ ] Added documentation to the new code and updated existing documentation where needed


ClaudiaMia (approved):
Looking very _green_ to me!~

PR #913, opened 2025-03-07
[FEATURE] Added timer functionality to locks

### LINKED ISSUE #849: Add timer lock
Add a new lock type and an asset using it that allows to set a time and then lock the item with the set time. 
Only after the timer runs out, the lock can be unlocked. That said, the person who locked the lock shall be able to unlock the lock even while the timer is still running.

While an advanced version could allow to define who is allowed to decrease and increase the timer, in the initial version this may not be needed, as the creator can at any time open the lock and lock it with a new time.

Livie53:
## References

resolves #849 

## About The Pull Request

Adds the necessary lock handling to support timer lock assets.

## Changelog

Authored by: Livie53

```md
Technical changes:
- Added timer functionality to lock handling to support timer locks
```

## Checklist

### DIFF
--- a/pandora-client-web/src/components/wardrobe/modules/wardrobeModuleLockSlot.tsx
+++ b/pandora-client-web/src/components/wardrobe/modules/wardrobeModuleLockSlot.tsx
@@ -231,16 +231,17 @@ function WardrobeLockSlotLocked({ target, item, moduleName, lock }: Omit<Wardrob
 
 		return lockedUntil - now;
 	}, [lock, now]);
+	const timerExpired = useMemo(() => (timeLeft == null || timeLeft <= 0), [timeLeft]);
 	const timerText = useMemo(() => {
-		if (timeLeft == null || timeLeft <= 0)
+		if (timerExpired)
 			return <>Remaining time: <i>Timer expired</i></>;
 
 		return (
 			<>
-				Remaining time: { FormatTimeInterval(timeLeft) }
+				Remaining time: { FormatTimeInterval(timeLeft ?? 0) }
 			</>
 		);
-	}, [timeLeft]);
+	}, [timeLeft, timerExpired]);
 
 	const [password, setPassword] = useState<string>('');
 	const [invalidPassword, setInvalidPassword] = useState<string | undefined>(undefined);
@@ -267,7 +268,9 @@ function WardrobeLockSlotLocked({ target, item, moduleName, lock }: Omit<Wardrob
 				clearLastPassword,
 			},
 		},
-	}), [clearLastPassword, currentAttempt, item, moduleName, password, target]);
+		// Add timerExpired to dependencies to force recalculation once timer runs out
+		// eslint-disable-next-line
+	}), [clearLastPassword, currentAttempt, item, moduleName, password, target, timerExpired]);
 
 	return (
 		<>
--- a/pandora-client-web/src/components/wardrobe/modules/wardrobeModuleLockSlot.tsx
+++ b/pandora-client-web/src/components/wardrobe/modules/wardrobeModuleLockSlot.tsx
@@ -269,7 +269,7 @@ function WardrobeLockSlotLocked({ target, item, moduleName, lock }: Omit<Wardrob
 			},
 		},
 		// Add timerExpired to dependencies to force recalculation once timer runs out
-		// eslint-disable-next-line
+		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}), [clearLastPassword, currentAttempt, item, moduleName, password, target, timerExpired]);
 
 	return (
--- a/pandora-client-web/src/components/wardrobe/modules/wardrobeModuleLockSlot.tsx
+++ b/pandora-client-web/src/components/wardrobe/modules/wardrobeModuleLockSlot.tsx
@@ -222,19 +222,22 @@ function WardrobeLockSlotLocked({ target, item, moduleName, lock }: Omit<Wardrob
 		const lockedData = lock.lockLogic.lockData.locked;
 		Assert(lockedData != null);
 
-		const { time, timerMinutes } = lockedData;
-		if (timerMinutes == null)
+		const { lockedUntil } = lockedData;
+		if (lockedUntil == null)
 			return null;
 
-		return time + (timerMinutes * 60_000) - now;
+		if (now >= lockedUntil)
+			return 0;
+
+		return lockedUntil - now;
 	}, [lock, now]);
 	const timerText = useMemo(() => {
 		if (timeLeft == null || timeLeft <= 0)
-			return 'Can be unlocked now';
+			return <>Remaining time: <i>Timer expired</i></>;
 
 		return (
 			<>
-				Can be unlocked in { FormatTimeInterval(timeLeft) }
+				Remaining time: { FormatTimeInterval(timeLeft) }
 			</>
 		);
 	}, [timeLeft]);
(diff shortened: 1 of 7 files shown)


pandora-client-web/src/components/wardrobe/modules/wardrobeModuleLockSlot.tsx
```diff
@@ -217,6 +218,26 @@ function WardrobeLockSlotLocked({ target, item, moduleName, lock }: Omit<Wardrob
...
+
+		return time + (timerMinutes * 60_000) - now;
+	}, [lock, now]);
+	const timerText = useMemo(() => {
+		if (timeLeft == null || timeLeft <= 0)
+			return 'Can be unlocked now';
+
+		return (
+			<>
+				Can be unlocked in { FormatTimeInterval(timeLeft) }
```
Jomshir98, line 237: [suggestion, optional] Just making it slightly nicer:
```suggestion
				Remaining time: { FormatTimeInterval(timeLeft) }
```

pandora-client-web/src/components/wardrobe/modules/wardrobeModuleLockSlot.tsx
```diff
@@ -217,6 +218,26 @@ function WardrobeLockSlotLocked({ target, item, moduleName, lock }: Omit<Wardrob
...
+
+		const { time, timerMinutes } = lockedData;
+		if (timerMinutes == null)
+			return null;
+
+		return time + (timerMinutes * 60_000) - now;
+	}, [lock, now]);
+	const timerText = useMemo(() => {
+		if (timeLeft == null || timeLeft <= 0)
+			return 'Can be unlocked now';
```
Jomshir98, line 233:
```suggestion
		if (timeLeft == null || timeLeft <= 0)
			return <>Remaining time: <i>Timer expired</i></>;
```

pandora-client-web/src/components/wardrobe/modules/wardrobeModuleLockSlot.tsx
```diff
@@ -309,9 +341,10 @@ function WardrobeLockSlotUnlocked({ target, item, moduleName, lock }: Omit<Wardr
 				password: currentAttempt != null ? undefined :
 					useOldPassword ? undefined :
 					(password || undefined),
+				timer: currentAttempt != null ? undefined : (timer || undefined),
```
Jomshir98, line 344: This breaks the "attempted action" logic. To see how it breaks, try putting on mittens and then try locking the lock.
This bit of code can be kept like this, but you should also do the following change:
In method `RedactSensitiveActionData` (in `pandora-common/src/gameLogic/actionLogic/actionUtils.ts`) set `timer` to `undefined` the same way `password` is handled in there.

pandora-client-web/src/components/wardrobe/modules/wardrobeModuleLockSlot.tsx
```diff
@@ -468,3 +512,58 @@ function PasswordInput({
...
+	const [hours, setHours] = useState<number>(0);
+	const [minutes, setMinutes] = useState<number>(0);
+
+	const maxHours = useMemo(() => {
+		return timer.maxMinutes / 60;
+	}, [timer]);
+
+	const maxMinutes = useMemo(() => {
+		return timer.maxMinutes < 59 ? timer.maxMinutes : 59;
+	}, [timer]);
```
Jomshir98, line 537: This approach technically works, but can behave in slightly unexpected ways at times. The fewer `useState`s you have, the better. In this case I would recommend passing the current `timer` value from `WardrobeLockSlotUnlocked` as a property and calculating `hours` and `minutes` from it.
Then `setHours` and `setMinutes` should be implemented in a way that they result in calling `onChange` with the new timer as a single value. (this then causes the state value to update in the parent component, propagating here, and causing the new calculation of `hours` and `minutes` from it)

pandora-client-web/src/components/wardrobe/modules/wardrobeModuleLockSlot.tsx
```diff
@@ -468,3 +512,58 @@ function PasswordInput({
...
+			</label>
+			<NumberInput
+				min={ 0 }
+				max={ maxHours }
+				step={ 1 }
+				value={ hours }
+				onChange={ setHours }
+				disabled={ pendingAttempt }
+			/>
+			<label>Hours</label>
```
Jomshir98, line 557: All labels should have unique value of `htmlFor`, matching the input they are for.
This means that optimally each `<NumberInput` should have unique id (e.g. `${id}-hours`, `${id}-mintues`) and the same should be set on its label - label describes what is the name of particular input, mainly important for assistive technologies (e.g. screen readers).

pandora-client-web/src/components/wardrobe/modules/wardrobeModuleLockSlot.tsx
```diff
@@ -468,3 +512,58 @@ function PasswordInput({
...
+			<NumberInput
+				min={ 0 }
+				max={ maxMinutes }
+				step={ 1 }
+				value={ minutes }
+				onChange={ setMinutes }
+				disabled={ pendingAttempt }
+			/>
+			<label>Minutes</label>
+		</Row>
```
Jomshir98, line 567: The layout of this is a bit weird. I would recommend doing the following instead:
```tsx
return (
	<Row className='WardrobeInputRow' alignY='center'>
		<label>
			Timer
		</label>
		<Row alignY='center'>
			<NumberInput
				id={ `${id}-hours` }
				...
			/>
			<label htmlFor={ `${id}-hours` }>Hours</label>
			{ ' : ' }
			<NumberInput
				id={ `${id}-minutes` }
				...
			/>
			<label htmlFor={ `${id}-minutes` }>Minutes</label>
			{ ' : ' }
			<NumberInput
				id={ `${id}-seconds` }
				...
			/>
			<label htmlFor={ `${id}-seconds` }>Seconds</label>
		</Row>
	</Row>
);
```

Jomshir98 (changes requested):
There are many smaller comments and changes, but overall decent work, especially so for first-time contribution to this part of codebase!

pandora-common/src/gameLogic/locks/lockSetup.ts
```diff
@@ -13,6 +13,11 @@ export interface LockSetup {
 		 */
 		format: 'numeric' | 'letters' | 'alphanumeric' | 'text';
 	};
+	/** Configuration to enable password on this lock */
+	timer?: {
+		/** Allowed minutes timer can be locked for */
+		maxMinutes: number;
```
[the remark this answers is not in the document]
Livie53: Changed

pandora-common/src/gameLogic/locks/lockLogic.ts
```diff
@@ -129,6 +130,13 @@ export class LockLogic {
 			};
 		}
 
+		if (this.lockSetup.timer != null && timer == null) {
+			return {
+				result: 'failed',
+				reason: 'noTimerSet',
+			};
+		}
```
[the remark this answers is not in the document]
Livie53: Left the `noTimerSet` for null/unset timers, but added the `invalidTimer` for mentioned invalid values

pandora-client-web/src/components/wardrobe/modules/wardrobeModuleLockSlot.tsx
Livie53: Fixed

Jomshir98 (approved):
LGTM!

PR #920, opened 2025-03-17
[REFACTOR] Split graphics data between runtime and editor, allow for multiple layer types

Jomshir98:
## References

_None_

## About The Pull Request

This rework focuses on two main things:
- Allowing for multiple types of layers.
  - Right now only two exist: A mesh-based image layer and mesh+image-based alphamasking layer.
  - The alphamasking layer was split from the standard image layer
- Splitting data that goes to editor and to normal Pandora. This is mainly so editor can have access to original data instead of optimized data that is used for Pandora
  - This also required for Editor to be able to build runtime data from its own data in-browser, which was another bit of refactoring.

## Changelog

```md
Asset creation:
- There are now types of layers for assets. At the moment two exist:
 - "Image layer" that is the standard layer you know, but option for adding alphamasks to it was removed
 - "Alpha image layer" that is similar to image layer, but it is only an alphamask
- Editor was reworked to work on original asset data instead of loading same data Pandora loads normally. Ideally you shouldn't notice this change, but it does remove the need to reverse any optimizations done during asset build, hopefully resulting in fewer quirks in Editor.

Technical changes:
- Reworked definition of data behind the scenes to allow for the above changes
- Editor now builds parts of asset graphics by itself, similar to what asset repository does (with the exception of optimizations and lower-quality image generation)
```

## Checklist


ClaudiaMia (approved):
LGTM~

PR #924, opened 2025-04-04
[FEATURE] Setting to disallow early timer unlock

Livie53:
## References

_None_

## About The Pull Request

Adds a checkbox for timer locks to disallow character locking it to unlock it before time running out. Checked by default, prompts on uncheck.

## Changelog

Authored by: Livie53
```md
Platform changes:
- Added ability to disallow unlocking a timer lock early even for the character that locked it
```

## Checklist


pandora-common/src/gameLogic/locks/lockData.ts
```diff
@@ -9,10 +9,17 @@ export const LockDataBundleSchema = z.object({
...
-		/** Time the timer on the lock will expire, if lock includes a timer */
-		lockedUntil: z.number().int().nonnegative().optional(),
 	}).optional(),
+	/** Data applicable only to locks with timer */
+	timer: z.object({
+		/** Time the timer on the lock will expire */
+		lockedUntil: z.number().int().nonnegative(),
+		/** Whether the character that locked it is allowed to unlock the timer early */
+		allowEarlyUnlock: z.boolean(),
+	}).optional(),
```
Jomshir98, line 19: This change breaks backwards compatibility without writing some additional migration code.
You should either write a migration code for this (which could be done using `z.preprocess` - look at `SpacePublicSettingSchema` for example).
Alternatively it could be done in a backward-compatible way in the first place by preserving the `lockedUntil` property as it is and introducing a new property along the lines of `disallowEarlyUnlock: z.literal(true).optional()` - where it missing is the current behavior.
I also wonder if the timer data shouldn't only be present in the `locked` section. Does it make sense for unlocked locks at all?

pandora-common/src/gameLogic/locks/lockLogic.ts
```diff
@@ -10,12 +10,19 @@ import type { AppearanceActionContext } from '../actionLogic/appearanceActions.t
 import type { LockDataBundle } from './lockData.ts';
 import type { LockSetup } from './lockSetup.ts';
 
+export const LockTimerOptionsSchema = z.object({
+	/** How long (in ms) the lock should be locked for, if it includes a timer. */
+	timer: z.number().int().nonnegative().optional(),
+	/** Whether the character that locked it is allowed to unlock the timer early, default true */
+	allowEarlyUnlock: z.boolean().optional(),
```
Jomshir98, line 17: Why make these properties optional at all, when the only thing done with the `timer` is missing is rejecting the request anyway? Making it non-optional removes need for checking that altogether and making `allowEarlyUnlock` non-optional simplifies its logic as well.

pandora-client-web/src/components/wardrobe/views/wardrobeLockLogic.tsx
```diff
@@ -236,6 +268,10 @@ export function WardrobeLockLogicUnlocked<TActionContext>({ lockLogic, ActionBut
 			{
 				lockLogic.lockSetup.timer ? (
 					<Column className='WardrobeLockTimer'>
+						<Row className='WardrobeInputRow'>
+							<label>Can be unlocked early by the locker</label>
+							<Checkbox checked={ timerAllowEarly } onChange={ setTimerAllowEarlyWithConfirm } />
```
Jomshir98, line 273: Checkbox should be before the text

Jomshir98, line 273: The text feels quite hard to understand.
Suggested alternative (as only the person locking it can control it anyway, as it isn't persistent while unlocked):
> The lock can be unlocked by you even the before timer runs out

pandora-client-web/src/components/wardrobe/views/wardrobeLockLogic.tsx
```diff
@@ -200,13 +210,35 @@ export function WardrobeLockLogicUnlocked<TActionContext>({ lockLogic, ActionBut
...
+		};
+	}, [timer, timerAllowEarly]);
+
+	const setTimerAllowEarlyWithConfirm = useCallback((newValue: boolean) => {
+		if (newValue) {
+			setTimerAllowEarly(newValue);
+			return;
+		}
+
+		confirm('Confirm disallow early unlock', <>Are you sure you want to disallow the locker to unlock before the timer runs out?</>)
```
Jomshir98, line 226: Similarly updating the text here:
```suggestion
		confirm('Confirm disallowing early unlock', <>Are you sure you want to prevent yourself from being able to unlock the lock before the timer runs out?</>)
```

Jomshir98 (changes requested):
A few improvement suggestions.

pandora-common/src/gameLogic/locks/lockLogic.ts
```diff
@@ -248,6 +256,7 @@ export class LockLogic {
 				name: player.appearance.character.name,
 				time: lockTime,
 				lockedUntil,
+				disallowEarlyUnlock: timerOptions?.allowEarlyUnlock,
```
Jomshir98, line 259: This looks wrong

Jomshir98 (changes requested):
The allow/disallow logic appears to be inverted. See the comment below.

pandora-client-web/src/components/wardrobe/views/wardrobeLockLogic.tsx
```diff
@@ -236,6 +269,10 @@ export function WardrobeLockLogicUnlocked<TActionContext>({ lockLogic, ActionBut
 			{
 				lockLogic.lockSetup.timer ? (
 					<Column className='WardrobeLockTimer'>
+						<Row className='WardrobeInputRow'>
+							<Checkbox checked={ timerAllowEarly } onChange={ setTimerAllowEarlyWithConfirm } />
+							<label>The lock can be unlocked by you even the before timer runs out</label>
```
Jomshir98, line 274: The checkbox should be given an id (with the help of `useId()`) and label should have `htmlFor` set to link it to the checkbox.
That will allow clicking the label text as well to toggle the checkbox, increasing usability. It also helps assistive tools describe the checkbox.

pandora-client-web/src/components/wardrobe/views/wardrobeLockLogic.tsx
```diff
@@ -150,6 +152,13 @@ export function WardrobeLockLogicLocked<TActionContext>({ lockLogic, ActionButto
 			{
 				lockLogic.lockSetup.timer ? (
 					<Column className='WardrobeLockTimer'>
+						{
+							lockLogic.lockData.locked?.disallowEarlyUnlock == null ? null : (
+								<Row className='WardrobeInputRow'>
+									Cannot be unlocked early
+								</Row>
+							)
```
Jomshir98, line 160: This should not show if `disallowEarlyUnlock === false` (it does currently).

Jomshir98 (approved):
Looking good now!

PR #926, opened 2025-04-05
[ADD] Card deck

SandrinePDR:
## References

_None_

## About The Pull Request

Adds a basic card game to Pandora that can be steered via the command line.  Depending on the community's feedback it may become more elaborate over time.

These are the new commands that can be used for the game
* create (create a new game)
* join (join an existing game)
* stop (end a game prematurely, dealer command)
* deal (deals a card to the room or a player, either open or face down, dealer command)
* check (check the cards you are holding)
* reveal (ends the game and shows the hands of all players, dealer command)

## Changelog

Authored by: Sandrine with a lot of  helpful comments by Clare

```md
Platform changes:
- Added a command driven simple card game (see above)

## Checklist

<!-- Checklist for you to make sure you didn't miss something -->

### CODE REVIEW

File: pandora-common/src/networking/client_shard.ts [hunk 1]
```diff
@@ -138,6 +139,13 @@ export const ClientShardSchema = {
 				type: z.literal('rps'), // Rock Paper Scissors
 				choice: z.enum(['rock', 'paper', 'scissors', 'show']),
 			}),
+			z.object({
+				type: z.literal('cards'), //CardGameCard game related messages
+				targetId: CharacterIdSchema.optional(),
+				msgOption: z.enum(['create', 'stop', 'join', 'deal', 'check', 'reveal']),
+				dealHidden: z.boolean().default(false).optional(),
+				card: CardGameCardSchema.optional(),
+			}),
```
Jomshir98, line 148: This type feels quite a bit like a mess. All the optional things are used in place of properly differentiating between message types.

I would recommend creating a new type in the utility/cards file, something along the lines of:
```ts
export const CardGameActionSchema = z.discriminatedUnion('action', [
    z.object({
        action: z.literal('create'),
    }),
    z.object({
        action: z.literal('deal'),
        dealHidden: z.boolean(),
        targetId: CharacterIdSchema.optional(),
    }),
    ... // And the rest
]);
```
And then the message as:
```ts
z.object({
    type: z.literal('cards'),
    action: CardGameActionSchema,
)}
```

This will make it a lot cleaner and define what options are valid for what action.
It will also make it much clearer when some options are not actually used (e.g. is the "card" option ever used?)

pandora-common/src/index.ts
```diff
@@ -14,4 +14,5 @@ export * from './security.ts';
 export * from './services/index.ts';
 export * from './space/index.ts';
 export * from './utility/index.ts';
+export * from './utility/cards.ts';
```
Jomshir98, line 17: This should be added in `pandora-common/src/utility/index.ts`, not `pandora-common/src/index.ts` to keep things clean.

pandora-server-shard/src/networking/manager_client.ts
```diff
@@ -414,6 +416,184 @@ export const ConnectionManagerClient = new class ConnectionManagerClient impleme
...
+										});
+									} else {
+										if (game.targetId) {
+											if (!game.dealHidden) {
+												space.handleActionMessage({
+													id: 'gamblingCardGameDealPlayerOpen',
+													character,
+													sendTo: space.cardGame.getPlayerIds(), // Send messages only to players
+													dictionary: {
+														'TARGET_CHARACTER': `${space.getCharacterById(game.targetId)?.name}`,
```
Jomshir98, line 487: If you can it is preferable to set `target` on the message (similar to `character`), as it lets Pandora have more control over that instead of just straight out string replacement through `dictionary` (e.g. we will be able to have the names in message interactive in the future similar to how item names are). Your dictionary replacement `TARGET_CHARACTER` already matches what setting `target` generates anyway.

pandora-server-shard/src/networking/manager_client.ts
```diff
@@ -414,6 +416,184 @@ export const ConnectionManagerClient = new class ConnectionManagerClient impleme
...
+														'CARD': card.toString(),
+													},
+												});
+											} else {
+												space.handleActionMessage({
+													id: 'gamblingCardGameDealPlayerSecret',
+													character,
+													sendTo: space.cardGame.getPlayerIds(), // Send messages only to players
+													dictionary: {
+														'TARGET_CHARACTER': `${space.getCharacterById(game.targetId)?.name}`,
```
Jomshir98, line 497: Same point about setting `target`.

pandora-server-shard/src/networking/manager_client.ts
```diff
@@ -414,6 +416,184 @@ export const ConnectionManagerClient = new class ConnectionManagerClient impleme
...
+									},
+								});
+								break;
+							case 'stop': {
+								// stops the current game
+								if (space.cardGame.isDealer(client.character.id)) {
+									space.handleActionMessage({
+										id: 'gamblingCardGameStopped',
+										character,
+										sendTo: space.cardGame.getPlayerIds(), // Send messages only to players
```
Jomshir98, line 442: Is there a reason to limit the message to players and not allow spectating to see what is going on? (valid for all messages that use this)

pandora-server-shard/src/networking/manager_client.ts
```diff
@@ -414,6 +416,184 @@ export const ConnectionManagerClient = new class ConnectionManagerClient impleme
...
+										character,
+									});
+								}
+								break;
+							}
+							case 'deal': {
+								if (space.cardGame.isDealer(client.character.id)) {
+									// Deals a card to a character or openly
+									const card = space.cardGame.dealTo(game.targetId);
+									if (!card) {
```
Jomshir98, line 474: Trying to deal to someone who isn't part of the game results in "There are no cards left in the deck" message, which is confusing.

pandora-client-web/src/ui/components/chat/commands.ts
```diff
@@ -321,6 +321,94 @@ export const COMMANDS: readonly IClientCommand<ICommandExecutionContextClient>[]
...
+							} else { //Deal card to room
+								shardConnector.sendMessage('gamblingAction', {
+									type: 'cards',
+									msgOption: 'deal',
+									dealHidden: false, // Cards to the room are always dealt openly
+								});
+							}
+							return true;
+						}),
+				},
```
Jomshir98, line 387: After playing with this command for a while, it feels rather confusing to use. My three main points are:
- It is not apparent that not using a target is a possibility (this comes from how `CommandStepOptional` works)
- The `/hidden` is actually most common, yet it is easy to accidentally not specify it
- It shows characters that are not part of the game (this unfortunately cannot be helped without syncing some data in advance)

I would suggest splitting this subcommand into three subcommands to take care of the first two steps: `dealTable`, `dealOpenly` and `dealHidden` (or simply `deal` as it is most common).

pandora-client-web/src/ui/components/chat/commands.ts
```diff
@@ -321,6 +321,94 @@ export const COMMANDS: readonly IClientCommand<ICommandExecutionContextClient>[]
...
+					description: 'Have a look at the cards that were dealt to you',
+					handler: ctx
+						.handler(({ shardConnector }) => {
+							shardConnector.sendMessage('gamblingAction', {
+								type: 'cards',
+								msgOption: 'check',
+							});
+							return true;
+						}),
+				},
```
Jomshir98, line 409: Playing around with this, I wonder if `check` should also show what cards are on the table (if any) for brevity.

Jomshir98: Potentially it could also show even more of the state of the game - which players are part of the game and how many cards they are holding, or even list of cards that were dealt face-up to them. (or this could be added as another option, such as `checkAll` or `status`, possibly even usable by those not part of the game).

pandora-server-shard/src/networking/manager_client.ts
```diff
@@ -414,6 +416,184 @@ export const ConnectionManagerClient = new class ConnectionManagerClient impleme
...
+									players.forEach((id) => space.handleActionMessage({
+										id: 'gamblingCardGameHandReveal',
+										character,
+										sendTo: players,
+										dictionary: {
+											'PLAYER': `${space.getCharacterById(id)?.name}`,
+											'HAND': `${space.cardGame?.getPlayerHand(id)}`,
+										},
+									}));
+									space.cardGame = null;
```
Jomshir98, line 566: [note] It honestly feels a bit weird that doing reveal stops the game, mainly because it forces all the players to re-register. But I think that that could be improved in future versions.

pandora-common/src/utility/cards.ts
```diff
@@ -0,0 +1,183 @@
...
+	suit: SuitSchema,
+	rank: RankSchema,
+}).transform((data) => new CardGameCard(data.suit, data.rank));
+
+/*
+export const CardGameCardDeckSchema = z.array(CardGameCardSchema).transform((cards) => {
+	const deck = new CardDeck(cards);
+	return deck;
+});
+*/
```
Jomshir98, line 183: This feels forgotten in here.

pandora-common/src/utility/cards.ts
```diff
@@ -0,0 +1,183 @@
...
+		this.players = [];
+		this.spaceHand = [];
+		if (!d) {
+			this.deck = new CardDeck();
+			this.deck.create();
+		} else {
+			this.deck = d;
+		}
+		this.dealer = creator;
+		this.players.push(new CardPlayer(this.dealer)); // No need to check, as we know the list of players is empty
```
Jomshir98, line 162: Nitpick: You could straight out do `this.players = [new CardPlayer(this.dealer)];`

pandora-common/src/utility/cards.ts
```diff
@@ -0,0 +1,183 @@
...
+	'\u2662', //diamonds
+	'\u2667', //clubs
+	'\u2664', //spades
+] as const;
+const ranks = ['2', '3', '4', '5', '6', '7', '8', '9', 'T', 'J', 'Q', 'K', 'A'] as const;
+
+// Define CardGameCard class
+export class CardGameCard {
+	public readonly suit: string;
+	public readonly rank: string;
```
Jomshir98, line 16: If you update the schemas like this:
```ts
const SuitSchema = z.enum(suits);
type CardGameSuit = z.infer<typeof SuitSchema>;

const RankSchema = z.enum(ranks);
type CardGameRank = z.infer<typeof RankSchema>;
```

Then you can update the card to be:
```ts
public readonly suit: CardGameSuit;
public readonly rank: CardGameRank;
```
Which will make it easier to spot mistakes in the future.

pandora-common/src/utility/cards.ts
```diff
@@ -0,0 +1,183 @@
...
+
+	public toString() {
+		return `${this.rank}${this.suit}`;
+	}
+
+	constructor(s?: string, r?: string) {
```
Jomshir98, line 22: Do the arguments need to be optional? If yes, then you can simplify the code by doing:
```ts
constructor(s: string = '\u2665', r: string = '2')
```

pandora-common/src/utility/cards.ts
```diff
@@ -0,0 +1,183 @@
...
+		this.suit = s ? s : '\u2665';
+		this.rank = r ? r : '2';
+	}
+}
+
+class CardArray extends Array<CardGameCard> {
+	public override toString(): string {
+		return `[ ${this.map((card) => card.toString()).join(', ')} ]`;
+	}
+}
```
Jomshir98, line 32: This class is never actually used. You use it only as type, but the `toString` will never be used at runtime.

pandora-common/src/utility/cards.ts
```diff
@@ -0,0 +1,183 @@
...
+
+	private addCard(c: CardGameCard) {
+		this.deck.push(c);
+	}
+
+	private shuffle() {
+		for (let i = this.deck.length - 1; i > 0; i--) {
+			const j = Math.floor(Math.random() * (i + 1));
+			[this.deck[i], this.deck[j]] = [this.deck[j], this.deck[i]];
+		}
```
Jomshir98, line 47: Nitpick: There already exists `ShuffleArray` (in `pandora-common/src/utility/misc.ts`) that you can use for this.

Jomshir98 (changes requested):
Hi! Sorry for taking quite long to review this - I was on a bit of a break from Pandora coding for a time.
Overall this is decent foundation and first version that can be built up upon more in the future. See the comments below for improvements I think make sense for the first version, still.

pandora-server-shard/src/networking/manager_client.ts
SandrinePDR: I wanted to avoid spamming the room in case of just a small group of people is actually playing.

pandora-server-shard/src/networking/manager_client.ts
```diff
@@ -414,6 +416,184 @@ export const ConnectionManagerClient = new class ConnectionManagerClient impleme
...
+									character,
+									sendTo: [client.character.id],
+									dictionary: {
+										'DEALER': `${space.getCharacterById(space.cardGame.getDealerId())?.name}`,
+									},
+								});
+								break;
+							case 'stop': {
+								// stops the current game
+								if (space.cardGame.isDealer(client.character.id)) {
```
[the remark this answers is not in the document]
SandrinePDR: Honestly, I thought about that as well. What would be your preference?

pandora-server-shard/src/networking/manager_client.ts
SandrinePDR: target: space.getCharacterById(game.action.targetId),

Seems to be wrong?

pandora-server-shard/src/networking/manager_client.ts
SandrinePDR: I added a safeguard. Would it be possible that only the current players are listed, when typing the command and using its auto-complete?

pandora-server-shard/src/spaces/space.ts
```diff
@@ -351,11 +354,34 @@ export abstract class Space extends ServerRoom<IShardClient> {
...
+							this.handleActionMessage({
+								id: 'gamblingCardGameStopped',
+								character: {
+									type: 'character',
+									id: character.id,
+								},
+								sendTo: [p],
+							});
+						}
+					});
```
Jomshir98, line 378: This can be simplified by passing character array to `sendTo` instead of calling it many times.

pandora-server-shard/src/networking/manager_client.ts
```diff
@@ -414,6 +416,236 @@ export const ConnectionManagerClient = new class ConnectionManagerClient impleme
...
+									dictionary: {
+										'HAND': space.cardGame.getPlayerHand(client.character.id),
+										'ISARE': (spaceHand === 'nothing' || spaceHand.length < 2) ? 'is' : 'are',
+										'TABLE': spaceHand,
+									},
+								});
+								break;
+							}
+							case 'reveal': {
+								if (space.cardGame.isDealer(client.character.id)) {
```
Jomshir98, line 594: Admin should be able to always stop the game, to handle cases like dealer going AFK.

Jomshir98 (changes requested):
Looks much better! The review contains the four points we already discussed and one additional forgotten comment.

PR #930, opened 2025-04-15
[FIX] Use lite preset for CSS nano

Jomshir98:
## References

_None_

## About The Pull Request

Turns out some transforms css nano does are not entirely safe. In this case a value `0%` was replaced by `0`, which normally makes sense, but it breaks the CSS in case of being used in `@property` declaration.
Instead of trying to hunt down the exact responsible optimization, I decided to simply tune down minification to the `lite` preset, which for the most part only minimizes whitespace.
This in turn increased CSS bundle size by roughly 2% (~2kb), which I think is more than acceptable price to pay for having fewer possible bugs.

## Changelog

```md
Fixes:
- Fixed the light animation in character selection screen jumping weirdly
```

## Checklist

### DIFF
--- a/pandora-client-web/package.json
+++ b/pandora-client-web/package.json
@@ -73,6 +73,7 @@
 		"cross-env": "7.0.3",
 		"css-loader": "7.1.2",
 		"css-minimizer-webpack-plugin": "7.0.2",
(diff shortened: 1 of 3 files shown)


ClaudiaMia (approved):
Good solution!

PR #936, opened 2025-04-28
[ADD] Delta updates for space state and testing framework for state and actions

### LINKED ISSUE #807: Performance: More optimized diff for appearance updates
At the moment whenever someone does appearance action we send rather big part of the current space's state. It might be beneficial to investigate better ways to create a diff that can be applied on client, reducing the size of updates.
This will be especially important if we move character's in-room positioning logic into this. It is already a bit problematic with bone positioning.

Jomshir98:
## References

resolves #807

_None_

## About The Pull Request

See the changelog or individual commits.

## Changelog

```md
Platform changes:
- Character modifier "Has item with attribute" condition now allows selecting from all existing attributes, not just those usable for asset preferences

Fixes:
- Fixed few bits of unclean code causing warnings in direct messages
- When you join a space, Pandora now always redirects you to the room view. This resolves weird behavior regarding space join links, especially those send through DMs.
- Fixed multiple updates being sent out during asset manager reload (which happens during local development or when new assets are deployed without server restart)

Technical changes:
- Reworked and heavily optimized how space state updates are sent to clients whenever some change happens to a character or room inventory. The results heavily depend on the specific space, but in most cases it results in the server sending 10-1000 times less data with each change. _Hopefully this won't create any new bugs_
- Reworked and heavily optimized how character modifier effects are send to the clients when they change. In the average this removes the need to send them entirely, which means infinitely times less data with each change compared to the prior implementation that sent them every single time.
- Added several automated tests for character and space state logic. _Mainly done as attempt to catch any new bugs from the above as early as possible_

Asset creation changes:
- Added remaining condition types to the automatic mesh layer type. This should make it usable for most current assets.
```

## Checklist


ClaudiaMia (approved):
LGTM~

PR #944, opened 2025-05-30
[ADD] Hearing impediment for long words

SandrinePDR:
## References

_None_

## About The Pull Request

Added a new option to the selective hearing modifier:
* Word length

This way long words, that are too complex for some characters can be scrambled with the set intensity.

## Changelog

Authored by: Sandrine

```md
Platform changes:
- Added a new configuration option "Words with at most this many letters can always be understood" to the already existing "Hearing: Selective hearing deprivation" character modifier

```

## Checklist

### DIFF
--- a/pandora-common/src/gameLogic/characterModifiers/modifierTypes/hearing_selective_deprivation.ts
+++ b/pandora-common/src/gameLogic/characterModifiers/modifierTypes/hearing_selective_deprivation.ts
@@ -54,21 +54,11 @@ You can, however, exclude characters or specific words from its effect or specif
 			},
 		},
 		wordLength: {
-			name: 'Distort words that are longer than a given value',
+			name: 'Distort only words that are longer than a given value',
 			type: 'number',
 			default: 8,
 			options: {
 				min: 0,
-			},
-		},
-		longWordDistortion: {
-			name: 'Intensity of long word scrambling',
-			type: 'number',
-			default: 9,
-			options: {
-				min: 0,
-				max: 10,
-				withSlider: true,
 			},
 		},
 	},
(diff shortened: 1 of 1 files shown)


ClaudiaMia (commented):
In line with my other comment, this should then be renamed in the opposite direction, we suggest:

"Words with at most this many letters can always be understood"

pandora-common/src/gameLogic/characterModifiers/modifierTypes/hearing_selective_deprivation.ts
```diff
@@ -53,6 +53,24 @@ You can, however, exclude characters or specific words from its effect or specif
 				withSlider: true,
 			},
 		},
+		wordLength: {
+			name: 'Distort words that are longer than a given value',
```
ClaudiaMia, line 57: In line with my other comment, this should then be renamed in the opposite direction, we suggest:

"Words with at most this many letters can always be understood"

pandora-common/src/gameLogic/characterModifiers/modifierTypes/hearing_selective_deprivation.ts
```diff
@@ -53,6 +53,24 @@ You can, however, exclude characters or specific words from its effect or specif
...
+			type: 'number',
+			default: 8,
```
ClaudiaMia, line 59: Default should be 0 as it is just an additional config setting for this rule and not the main thing.

ClaudiaMia (changes requested):

pandora-common/src/gameLogic/characterModifiers/modifierTypes/hearing_selective_deprivation.ts
ClaudiaMia: Instead, if you want to highlight the usage of the new config item for a doll hearing purpose with a default like the doll talk mod has it, you could add a preconfigured template to this mod. See file `characterModifierTemplates` to see examples.

ClaudiaMia (approved):
Thank you~

PR #947, opened 2025-06-06
[FEATURE] Legs split

Jomshir98:
## References

closes #622

## About The Pull Request

Splits left and right legs and soles. The order and combinations were chosen to keep the most compatibility with current assets, especially skirts.

## Changelog

```md
Platform changes:
- Graphics for the legs were split into left and right leg. This should fix most graphical issues there were with legs overlapping. As Pandora already has quite a few assets, it was impossible to test all combinations during testing. Please let us know if you notice any issues with legs after this update!
- Added an posing option to choose whether left or right leg should be on top of the other one.

Asset creation changes:
- Added a button to duplicate selected layer to Editor, as it is quite common to want another layer with very similar settings to an existing one
- Added a VSCode launch config for launching Editor directly.
- Reordered preview cutter inputs in Editor to match the order we usually write preview generation comments in asset files.
```

## Checklist


ClaudiaMia (approved):
LGTM!

PR #961, opened 2025-07-10
[FEATURE] Normals-based shading and specular lighting

Jomshir98:
## References

_None_

## About The Pull Request

This PR experiments with new way to dynamically shade and highlight assets that have normal maps (which are easy to get for assets generated from 3D models). It also includes a few fixes and Editor improvements.

## Changelog

```md
Platform changes:
- Added a new, experimental method of drawing assets that generates shading and highlights dynamically based on character's angle and pose. All assets added with this update are based on this new drawing. Note, that we do expect more improvements to this in the future.
- Added support for antialiasing and enabled it by default.
  - Note: Antialiasing conflicts with our current implementation of alphamasks, so they cannot be used at the same time
  - Note: Our antialiasing support relies fully on your browser and as this is an optional extension to drawing, it might not do anything depending on your browser and OS.

Fixes:
- Fixed that [Tab] would focus invisible elements, such as status selectors or notification menu while they are closed.
- Removed chat messages on module change altogether, as only two assets were using those and both of them were broken in some cases. We have plans for a more generic replacement mechanic in the future.

Asset creation changes:
- Asset Editor now supports specifying normal maps and associated metadata for automatic image layer. For manual image layer this isn't yet supported.
- Asset Editor Automesh Layer has a new feature to auto-fill images, if they are in the format `graphicslayer_variable1_variable2.png` to cut down on the most annoying part of adding layers.
```

## Checklist

### DIFF
--- a/pandora-client-web/src/assets/assetGraphicsCalculations.ts
+++ b/pandora-client-web/src/assets/assetGraphicsCalculations.ts
@@ -29,7 +29,7 @@ export function useLayerImageSource(
 	evaluator: AppearanceConditionEvaluator,
 	{ image: scalingBaseimage, scaling }: Pick<Immutable<Extract<GraphicsLayer, { type: 'mesh'; }>>, 'image' | 'scaling'>,
 	item: Item | null,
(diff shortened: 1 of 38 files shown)


ClaudiaMia (approved):
LGTM!~

PR #964, opened 2025-07-17
[FEATURE] Add exception to doll talk modifier

SandrinePDR:
## References

_None_

## About The Pull Request

Added an exception list to the doll talk modifier, so that these words can always be used, regardless of their length.

## Changelog

Authored by: Sandrine

```md
Platform changes:
- Added an exception list to the doll talk modifier. Words on this list can always be used, regardless of their length
```

## Checklist


pandora-common/src/gameLogic/characterModifiers/modifierTypes/speech_doll_talk.ts
```diff
@@ -27,16 +27,28 @@ _Note: Setting '0' means this part is not limited._
...
+				maxEntryLength: 24,
+				matchEntry: /^\p{L}+$/igu,
+			},
+		},
 	},
 	checkChatMessage(config, message) {
 		if (message.type !== 'chat')
 			return { result: 'ok' };
 
+		const allowedWordSet = new Set(config.wordAllowlist.map((w) => w.toLocaleLowerCase()));
```
Jomshir98, line 45: Please don't use `Locale*` methods in code that runs both on server and the client. While it is rare for the behavior to differ between locales, there certainly are cases where that happens - in that case the code would behave differently on client and server, which could lead to issues.
Use `toLowerCase` instead.

Jomshir98 (changes requested):

ClaudiaMia (approved):
Allowing to add up to 100 words as exception feels a bit excessive to me, as that may then no longer feel like a doll talk condition any more, depending on how many exceptions are used in one sentence.

But then again, if someone really feels the need for so many exceptions, I guess there is no real reason to forbid it, as it seems unlikely that many users will add so many words to this modification and use up a lot of server space.

PR #980, opened 2025-08-15
[FIX] Crash fix in render when mesh changes size and prevent non-admin moving deployed room devices

Jomshir98:
## References

_None_

## About The Pull Request

See changelog

## Changelog

```md

Fixes:
- Moving a deployed room device between room inventories now properly checks that you are a space admin
- Fixed a crash that occasionally occurred when moving from a 3D-Box room with large walls directly to another 3D-Box room with either small walls or walls with big tiles and then triggered a new frame render, e.g. by zooming in
- Fixed an error that occasionally occurred when moving from a 3D-Box room with small walls directly to another 3D-Box room with either large walls or walls with small tiles, that caused the walls in question to get stuck on the screen instead of moving with the camera

Technical changes:
- Improved handling of errors during graphics rendering - they will no longer crash the whole of Pandora, but only the relevant graphics view
```

## Checklist

### DIFF
--- a/pandora-client-web/src/editor/graphics/export/imageExporter.ts
+++ b/pandora-client-web/src/editor/graphics/export/imageExporter.ts
@@ -62,6 +62,7 @@ class TextureCutter extends Container {
 			uvs: uv,
 			indices: triangles,
 		});
+		geometry.batchMode = 'no-batch';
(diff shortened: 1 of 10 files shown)


ClaudiaMia (approved):
<3

PR #992, opened 2025-09-14
[FIX] Several fixes and new priority layer

Jomshir98:
## References

_None_

## Changelog

```md
Fixes:
- Assets which set default limit to anything other than "normal" in their asset definition do not take priority over asset attributes limits (so if item defaults to "maybe" but has attribute that you manually set to "prevent", it will now correctly resolve to "prevent" item limit)
- Added support for colorization migration, so changing colorization definition will not erase everyone's eye color in the future. This will prevent the color reset for characters that didn't load since the last update, but it will not retroactively restore colorization that has already been lost. We apologize for the inconvenience.
- Fixed several issues related to Zod migration in the last update that could cause some account or character data to get lost in some cases. Specifically the following losses were detected upon further investigation:
  - Six accounts lost their account description. Those will be restored from backup soon.
  - Two spaces lost their "allowed users" configuration. Those will be restored from backup soon.
  - One character (Claudia) lost their item limit settings.
- Added additional checks that should help us discover similar errors to the one above much sooner in the future
- Fixed a minor issue that prevented client from reconnecting to the server for the first ~15 seconds after a server restart

Technical changes:
- Database migration has been enabled by default to avoid many weird issues originating from having created a development environment in the past and not having updated it since.

Asset creation changes:
- Added a new `ABOVE_BACK_HAIR` priority layer, that will always be below both arms (if they are behind the back), but above back hair (from front view)
- Setting `disablePoseOffset` on a room device will now correctly prevent the pose from affecting character scale as well (so sitting will no longer cause character to get smaller)
```

## Checklist


ClaudiaMia (approved):
LGTM!

PR #996, opened 2025-09-19
[ADD] Movement and chat improvements, asset info

### LINKED ISSUE #526: Character disconnect should reset character's typing status
Right now if character disconnects while typing, their status can be stuck as "Name is typing" for others in the room.
The following changes needs to be done:
- When a client disconnects from character (the character doesn't necessarily need to go offline), then their typing status should be set to "none" by the shard

Following changes are nice-to-have:
- When a client reconnects it should receive the current typing status from all characters currently in the room (otherwise it might have outdated status). This should be possible as the shard's space already saves the current status for everyone.
- When a client reconnects, the client should resend the current typing status (or at least reset it) so it is properly propagated back to the shard.

Jomshir98:
## References

fixes #526

## Changelog

Authored by: Clare & Claudia

```md
Platform changes:
- Character following another one in the "Leash" mode can now be moved manually within the leash distance.
- Typing indicator now properly works with chat commands (such as `/me`, `/ooc`, ...).
- Chat commands that perform an action (such as posing, or follow/lead) will wait until the action is completely and will properly report problems that happen during that.
- Added asset creator credits to the "Create new item" menu. You can now also search for an asset of a specific contributor by using their name. 
- Updated style of the section for showing item's custom name and description.
- Added an option to enter custom item name and description directly during item creation.
- Added a new option to open information about an asset to the above-mentioned section, with list of asset creators and a link to asset's source code and images.

Fixes:
- Attempting to complete a slow action sooner than should be possible will no longer cancel it.
- It is no longer possible to accidentally break following by moving to another room inside space while following a character.
- A "Join-me" invite can no longer be created if it wouldn't be usable anyway (such as by a non-admin user into a private space).
- Typing indicator will now properly clear for character that disconnects, instead of being stuck until reload.
- Active typing indicators will now be shown for newly joining/reconnecting characters, instead of having to wait for their state to change.
- Disconnected client will now properly sync typing indicator to server after reconnect.

Technical changes:
- Chat commands now support arbitrary `async` executors.
- Fixed an issue where Shard was sending useless updates, due to trying to send server-only data to client when they changed.
```

## Checklist


ClaudiaMia (approved):
LGTM!~

PR #1030, opened 2025-11-28
[CHANGE] Several smaller changs and fixes

Jomshir98:
## References

_None_

## Changelog

```md
Platform changes:
- Attempting to move a character you cannot move without a permission prompt will now prompt you for confirmation whether you actually want to send a permission prompt.
- When importing a room template, non-deployed room devices now also have their position imported. Note, that the position was already being exported, so there is no need to re-create existing templates.
- Right-clicking a chat message twice (while Pandora's context menu is visible after first click) will now open the default right-click menu, allowing you to use common browser actions.
- Hovering a message time in chat will now display more precise time and a character that sent the message (especially useful for `**`-type messages).
- Added subtle shadow effect to character names in chat and other places where they are colored, to increase their legibility no matter the color.
- Pinned item dialogs now keep their position, size, and state while switching spaces.
- Blocking an account will now apply silhouette effect on its character by default. This is configurable in the settings and can be overridden temporarily using the "Temporarily hide" feature.
- Added an icon to space's character list for characters of a blocked account.
- Added a badge on room configuration settings button indicating when room has some non-default settings.
- Improved layout of space configuration room grid to better show room names, especially with many rooms present.
- Improved room name handling on the room minimap.
- Small tutorial and wiki changes

Fixes:
- Draw order of characters and room devices with the same Y coordinate is now well-defined, preventing issues such as two people seeing character and room device overlaying each other differently.
- Fixed an issue with character names and room devices, which could cause a character name to show under a room device without the character actually being in it, if the character was previously forcefully removed from the space while in a device and later returned.
- Fixed that a leaving character could be shown as `[UNKNOWN]` in a leave message, if the character has left the space already and later re-joined without the space getting unloaded meanwhile, if at least one different character in the space left at least a minute after the original leave.
- Fixed that long item names inside the wardrobe could push the icon and quick buttons outside of the item line.
- Fixed icon misalignment in wardrobe on iOS (I hope - I don't have an Apple device myself).
- Fixed an issue in space invite creation dialog that could break it visually on small screens.
- Fixed context menu displaying above modal dialogs.

Technical changes:
- Added a new `clean` script to clean up all build and testing artifacts in the whole repository
- Several internal changes and cleanup related to draw ordering in room.
- Fixed a subtle not-user-facing bug that was causing personal spaces to reload each time there was any shard update.
```

## Checklist

### DIFF
--- a/pandora-client-web/src/graphics/room/roomCharacter.tsx
+++ b/pandora-client-web/src/graphics/room/roomCharacter.tsx
@@ -303,6 +303,7 @@ export const RoomCharacterInteractive = memo(function RoomCharacterInteractive({
 					// Skip, we already asked recently
 					return false;
 				}
+				setPositionErrorCooldown.current = Date.now() + LIVE_UPDATE_ERROR_THROTTLE;
 
 				return confirm('Move permission request', 'You do not currently have permission to move this character. Send permission prompt?');
 			})


ClaudiaMia (approved):
LGTM!

PR #1035, opened 2025-12-04
[CHANGE] Replaced grphics with Unicode in asset tab

SandrinePDR:
## References

_None_

## About The Pull Request

Replaced built-in graphics with Unicode characters so the editor elements are usable universally 

## Changelog

Authored by: Sandrine

```md
Platform changes:
- Changed graphics to Unicode in the editor's asset tab
- Added a new CSS class to keep the Unicode buttons from 'jumping'
```

## Checklist

- [X] I understand this patch is submitted under the [Pandora's Contributor Agreement](https://github.com/Project-Pandora-Game/pandora/blob/master/contributor-license-agreement.md)

### DIFF
--- a/pandora-client-web/src/components/common/button/button.scss
+++ b/pandora-client-web/src/components/common/button/button.scss
@@ -62,6 +62,18 @@
 		padding: 0.1em 0.3em;
 	}
 
+	&.slim-fixed {
+		width: 1.8em;
+		height: 1.8em;
+		padding: 0;
+		display: inline-flex;
+		align-items: center;
+		justify-content: center;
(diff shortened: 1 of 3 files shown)


Jomshir98:
> Replaced built-in graphics with Unicode characters so the editor elements are usable universally

Not quite sure what do you mean here? Those already were unicode characters, just written directly in code instead of using escape sequence. Why would this change be needed? Having them in code as characters makes it more clear what they are for anyone looking through, which escape sequences definitely don't do.

SandrinePDR:
Because currently the editor looks like this for me
image

PR #1036, opened 2025-12-06
[ADD] Forced colors support and few small improvements

Jomshir98:
## References

_None_

## Changelog

```md
Platform changes:
- Added support for "System colors" and "Forced colors" browser features. If these are enabled (first through Pandora's settings, second depending on browser and OS accessibility options, most notably the "High Contrast Windows accessibility feature"), then Pandora's UI colors are almost entirely decided by your browser and OS settings.
- Several small changes, fixes and cleanup of various bits of UI styling throughout Pandora
- Settings screen now has a new way for opening individual setting pages, which should make it more easy to navigate settings on both desktop and mobile.
- Added ability to view complete action data to the action log messages.
- Made item names click-able in the action log messages, behaving the same as item names in standard action messages.
- Added account display name and id to all character join/leave server messages.
- Added account display name to the character info in the "Room" tab.
- Added current character count to the "Room" tab, above character list.

Technical changes:
- Space's "development" settings can now be changed even after space creation, if "development" feature was enabled during creation and the account has developer role.
- Several changes to our client build pipeline:
  - Dropped `postcss-preset-env` in favor of using only `autoprefixer` for CSS. `autoprefixer` is part of `postcss-preset-env` and is very useful, but other transforms present in `postcss-preset-env` tend to have high degree of unintended side-effects that are hard to debug.
  - Dropped `cssnano` altogether, as compared to the minification SCSS does by itself, it saves only additional 0.5% for our project - not worth the overhead it has, both dependecy-wise and build-time-wise.
  - Reduces our browser-list settings to only actively support browser versions released in the last 2 years instead of 3, as emerging "widely available" mark is set at 30 months of support.
- Room controls now have an error boundary, which avoids crashing the whole page if there is an error in this complex piece of Pandora.
```

## Checklist

### DIFF
--- a/pandora-client-web/src/components/gameContext/interfaceSettingsProvider.tsx
+++ b/pandora-client-web/src/components/gameContext/interfaceSettingsProvider.tsx
@@ -3,17 +3,17 @@ import { useEffect } from 'react';
 import { useAccountSettings } from '../../services/accountLogic/accountManagerHooks.ts';
 
 export function InterfaceSettingsProvider(): null {
-	const { interfaceAccentColor, forceSystemColors } = useAccountSettings();
+	const { interfaceAccentColor, accessibilityForceSystemColors } = useAccountSettings();
 
 	useEffect(() => {
 		document.body.style.setProperty('--pandora-accent-color', interfaceAccentColor);
 
-		if (forceSystemColors) {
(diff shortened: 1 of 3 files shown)


ClaudiaMia (approved):
LGTM!

PR #1105, opened 2026-05-10
[FEATURE] Dropped lock UI and lock template data, update pnpm to v11

Jomshir98:
## References

_None_

## Changelog

```md
Platform changes:
- Saving locks as saved items now saves the list of registered fingerprints, allowing you to spawn the lock with its saved list directly. This works for both standalone saved locks and locks saved as part of another saved item.
- Added ability to interact with locks that are not in a lock slot (e.g. in a room inventory or stored in some container). You can now change the lock's settings and even lock and unlock them, even if that won't do much for you.
- Added a new chat command `/wardrobe`, which can be used to open wardrobe of any character or room in the current space.
- While creating a new item (or item from saved item template), typed modules now highlight the default variants.

Technical changes:
- Reworked some logic around routing lock actions, simplifying how interacting with locks works behind the scenes and no longer requiring lock slots to interact with locks.
- Some dependency updates.
```

## Checklist

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,6 +1,0 @@
-update-notifier=false
-strict-peer-dependencies=true
-verify-deps-before-run=error
-inject-workspace-packages=false
-force-legacy-deploy=true
-node-options="--no-warnings=ExperimentalWarning --experimental-vm-modules"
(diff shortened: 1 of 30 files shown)


ClaudiaMia (approved):
Great work!

PR #1106, opened 2026-05-16
[FIX] Cleaned stale room device slots

### LINKED ISSUE #714: Clear kicked/banned players from device slots automatically

Sekkmer:
## References

fixes #714

## About The Pull Request

This fixes stale room device slot occupancy after a character leaves, is kicked, or is removed from a space.

The Directory now sends explicit character-removal cleanup intents to shards. Shards use those intents to clear room device slots for real character removals before a same-update space unload, while ordinary shard unloads and migrations no longer look like every character left.

Loaded spaces also reconcile invalid room device slot occupancy, so already-stale state is cleaned when the space is loaded or updated without requiring a database migration.

Validated locally with focused state, Directory protocol, and shard update tests.

## Changelog

Authored by: Sekkmer

```md
Fixes:
- Fixed room device slots staying occupied after the occupying character left or was removed from the space.
```

## Checklist


Jomshir98 (approved):
Looks good to me! Thank you.

PR #1107, opened 2026-05-18
[FEATURE] Add passkey sign-in and sudo mode

Sekkmer:
## References

_None_

## About The Pull Request

This adds passkey support for Pandora accounts using WebAuthn with resident/discoverable credentials and the PRF extension. Users can register passkeys, sign in without entering a username, and unlock their direct message key client-side without sending the wrapping secret to the server.

The PR also adds a connection-scoped account security unlock flow for privileged account actions. Password change, passkey management, and character deletion now request this temporary security unlock when needed; the unlock can be completed with the current password or a registered passkey.

Passkey sign-in is account-discoverable: the browser or passkey manager presents the available Pandora passkeys and Pandora resolves the selected account from the returned credential id. The login start response stays opaque and does not return account-specific credential lists before authentication. Browsers that support conditional mediation can show the passkey account picker from the login form; the explicit passkey button remains as a fallback.

The change includes:

- Resident passkey registration and username-less passkey sign-in from the login screen
- Conditional mediation support with `autocomplete="username webauthn"` for browser-managed account selection
- Account security unlock for privileged actions, backed by current-password or passkey verification
- Settings > Security flows for changing password and managing passkeys after security unlock
- Character deletion flow updated to request security unlock instead of requiring a separate password path
- Passkey-based direct message key wrapping and unlock, including remembered-login recovery for passkey sessions
- Server-side WebAuthn challenge validation, RP ID hash validation, origin checks, signature verification, sign-counter tracking, and challenge expiry
- Account lookup by passkey credential id, backed by a sparse unique Mongo index
- Password reset cleanup that removes passkeys when the saved direct message key is reset
- Browser credential cleanup for rejected passkey credentials when supported
- Account security wiki documentation
- Mock dev test account seeding for local testing

Deployment configuration:

For production, configure passkeys to use the shared RP ID:

```env
PASSKEY_RP_ID="project-pandora.com"
PASSKEY_ALLOWED_ORIGINS="https://project-pandora.com,https://www.project-pandora.com,https://fallback.project-pandora.com,https://www.fallback.project-pandora.com"
```

`PASSKEY_RP_ID` should stay `project-pandora.com` so credentials work across the main site and subdomains such as fallback. The server validates exact origins, so every browser origin that can serve the app needs to be listed in `PASSKEY_ALLOWED_ORIGINS`.

Screenshots captured locally:

- Sign-in passkey UI:
Screenshot_20260518_170142

- Account security passkey settings and passkey password change:
Screenshot_20260518_170319

Tested locally with a YubiKey 5C NFC device on Firefox and Chrome. The rebuilt stack was also validated with focused passkey tests, lint, type-check, and the full workspace test suite.

## Changelog

Authored by: Sekkmer

```md
Platform changes:
- Added passkey support for signing in with compatible browsers and security keys/password managers.
- Added security settings for managing registered passkeys.
- Privileged actions such as password changes, passkey management, or character deletion now have unified identity check before allowing you to do them.

Technical changes:
- Implemented WebAuthn passkey validation and passkey-based direct message key wrapping.
- Implemented resident passkey login with browser-managed account selection and opaque pre-authentication responses.
- Implemented password/passkey "sudo" re-authentication flow shared by account security operations.
- Added account security documentation and local development test-account improvements.
```

## Checklist


pandora-common/src/inputLimits.ts
```diff
@@ -25,6 +24,36 @@ export const LIMIT_MAIL_LENGTH = 256;
 /** Space ownership limit for an account */
 export const LIMIT_SPACE_OWNED_COUNT = 5;
 
+/** Maximum number of passkeys registered to one account */
+export const LIMIT_ACCOUNT_PASSKEY_COUNT = 5;
+/** Maximum length of a user-visible passkey name */
+export const LIMIT_ACCOUNT_PASSKEY_NAME_LENGTH = 80;
+/** Maximum length of a passkey credential id */
+export const LIMIT_ACCOUNT_PASSKEY_CREDENTIAL_ID_LENGTH = 2048;
```
Jomshir98: I wonder if limits that aren't user-facing should be in this file. This file is meant to organize limits so they can be tweaked if there is a need, but things like credential id might make more sense written at its specific type directly, as there is a technical reason for this, not management one.

pandora-common/src/networking/client_directory.ts
```diff
@@ -154,6 +154,32 @@ export const ClientDirectorySchema = {
...
 	},
+	passkeyLoginStart: {
+		request: z.object({
+			secondFactor: SecondFactorDataSchema.optional(),
+		}),
+		response: ZodCast<{
+			result: 'ok';
+			rpId: string;
+			challenge: string;
+			credentials: { id: string; type: 'public-key'; transports?: string[]; }[];
```
Jomshir98: It seems to me the "credentials" field is always empty? Is this future-proofing or is it forgotten remnant?

pandora-common/src/networking/client_directory.ts
```diff
@@ -154,6 +154,32 @@ export const ClientDirectorySchema = {
...
+			prfSalt: string;
+		} | { result: 'unknownCredentials'; } | SecondFactorResponse>(),
```
Jomshir98: Start can never return `unknownCredentials`, as far as I can see.

pandora-common/src/networking/client_directory.ts
```diff
@@ -165,6 +191,28 @@ export const ClientDirectorySchema = {
 		}),
 		response: ZodCast<{ result: 'ok' | 'invalidPassword'; }>(),
 	},
+	passkeyPasswordChangeStart: {
```
Jomshir98: Idea: We have quite a few endpoints that require current password (e.g. password change, passkey change, deleting character). The character deletion can't even be done using passkeys, as far as I can see.
How about introducing a "sudo" mode, which would be a flag on connection that resets after 5 minutes or when connection closes (this includes short DCs, but tough luck if you really hit one of those during that action, you just have to do it again).
And then we could have one universal endpoint for enabling sudo, and all privileged endpoints like password/passkey change, or character deletion could just check this flag instead or return "sudoRequired", which would trigger sudo auth dialog (similar to second factor) before retrying the request.

pandora-common/src/networking/client_directory.ts
```diff
@@ -154,6 +154,32 @@ export const ClientDirectorySchema = {
...
+			credentialId: AccountPasskeyCredentialIdSchema,
+			clientDataJSON: AccountPasskeyClientDataSchema,
+			authenticatorData: AccountPasskeyAuthenticatorDataSchema,
+			signature: AccountPasskeySignatureSchema,
+		}),
+		response: ZodCast<{ result: 'unknownCredentials'; } | {
+			result: 'ok';
+			token: { value: string; expires: number; };
+			account: IDirectoryAccountInfo;
+			cryptoKey: IAccountCryptoKey;
```
Jomshir98, line 185: Nice to have: It would be great to return list of passkeys with successful login (both passkey-based and normal login). Then you can send that passkey list to [`signalAllAcceptedCredentials`](https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential/signalAllAcceptedCredentials_static), which informs the browser/password manager to delete all resident passkeys for this user that aren't valid anymore.

Jomshir98, line 185: Nice to have: Similar to signalling all accepted credentials, there is also [`signalCurrentUserDetails`](https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential/signalCurrentUserDetails_static), which can be used to automatically update details on the passkeys, such as saved user's display name.
This could be used to update the saved display name to the configured account's display name whenever user logs in, to make the passkeys look nicer in the PM.

pandora-common/src/networking/client_directory.ts
```diff
@@ -154,6 +154,32 @@ export const ClientDirectorySchema = {
...
+		} | { result: 'unknownCredentials'; } | SecondFactorResponse>(),
+	},
+	passkeyLoginFinish: {
+		request: z.object({
+			credentialId: AccountPasskeyCredentialIdSchema,
+			clientDataJSON: AccountPasskeyClientDataSchema,
+			authenticatorData: AccountPasskeyAuthenticatorDataSchema,
+			signature: AccountPasskeySignatureSchema,
+		}),
+		response: ZodCast<{ result: 'unknownCredentials'; } | {
```
Jomshir98: Nice to have: If login results in `unknownCredentials`, you can use [`signalUnknownCredential`](https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential/signalUnknownCredential_static) to inform the PM that the passkey is no longer valid, which will cause it to be deleted to stop confusing the user.

pandora-common/src/networking/client_directory.ts
```diff
@@ -238,6 +286,52 @@ export const ClientDirectorySchema = {
...
+	},
+	passkeyRegisterFinish: {
+		request: z.object({
+			name: AccountPasskeyNameSchema,
+			credentialId: AccountPasskeyCredentialIdSchema,
+			publicKey: AccountPasskeyPublicKeySchema,
+			clientDataJSON: AccountPasskeyClientDataSchema,
+			authenticatorData: AccountPasskeyAuthenticatorDataSchema,
+			transports: AccountPasskeyTransportSchema.array().max(LIMIT_ACCOUNT_PASSKEY_TRANSPORT_COUNT).optional(),
+			prfSalt: AccountPasskeyPrfSaltSchema,
```
Jomshir98: Should PRF salt really be saved on passkey? For resident passkeys to work properly, it needs to be same for everyone using same rpId anyway...

pandora-client-web/src/crypto/passkey.ts
```diff
@@ -0,0 +1,246 @@
...
+				residentKey: 'required',
+				userVerification: 'required',
+			},
+			excludeCredentials: start.excludeCredentials.map((excludedCredential) => ({
+				...excludedCredential,
+				id: Base64UrlToArray(excludedCredential.id),
+				transports: excludedCredential.transports as AuthenticatorTransport[] | undefined,
+			})),
+			extensions: {
+				prf: {},
```
Jomshir98: Spec allows optionally evaluating salt on creation already. Most notably passing salt on creation is guaranteed not to break authenticators that don't support that.
This means you can do eval during creation optimistically, avoiding the second prompt if the authenticator supports this, only doing the later `GetPrfSecretForNewCredential` if you didn't get the result already during creation.

pandora-client-web/src/crypto/passkey.ts
```diff
@@ -0,0 +1,246 @@
...
+		return authenticatorData;
+	}
+
+	const attestation = new Uint8Array(response.attestationObject);
+	const authDataIndex = FindCborTextKey(attestation, 'authData');
+	if (authDataIndex < 0)
+		throw new Error('Unable to find authenticator data');
+
+	const marker = attestation[authDataIndex];
+	if (marker === 0x58) {
```
Jomshir98: Could there please be some comments or even better Spec link with these random constants? (applies on Directory too).
Also what do we use attestation for? Is it part of the spec, or optional? Do we even want attestation, or do we want to give people the most freedom with what they use for passkeys?

pandora-client-web/src/services/accountLogic/accountManager.ts
```diff
@@ -102,6 +114,52 @@ class AccountManager extends Service<AccountManagerServiceConfig> implements IAc
...
+		}
+
+		directoryConnector.authToken.value = { ...result.token, username: result.account.username };
+		await this.handleAccountChange({
+			account: result.account,
+			character: null,
+			passkeyUnlock: {
+				cryptoKey: result.cryptoKey,
+				wrappingSecret: assertion.wrappingSecret,
+			},
```
Jomshir98, line 168: This is unfortunately not enough. There is currently a bug:
- Log in with passkey
- Refresh page
- Open DMs -> It fails, requiring you to re-log.
This is a blocker bug in my eyes.

Do we store the wrapping key somewhere? If yes, then we might need to add a note either on the login token or somewhere in the token flow which passkey was used for login, so client receives the correct wrapped DM key during that flow, letting it decrypt it with the stored secret.
I am wary of just storing the decrypted secret, as that could have subtle implications in cases like lost device. In my opinion E2E key should always require cooperation of both server and client to get unlocked, even for devices with remembered login.

pandora-server-directory/src/account/accountSecure.ts
```diff
@@ -131,6 +131,29 @@ export default class AccountSecure {
 		return true;
 	}
 
+	public async changePasswordWithPasskey(passwordNew: string, cryptoKey: IAccountCryptoKey): Promise<'ok' | 'invalidCryptoKey'> {
```
Jomshir98: This made me wonder... should password reset (not just change) wipe all stored passkeys? Password reset already regenerates the E2E key, so that is likely to cause bugs otherwise.

Jomshir98 (changes requested):
First round of review.

pandora-client-web/src/crypto/passkey.ts
Sekkmer: that failed for me and could not get it do a single auth but might give it a go

pandora-server-directory/src/networking/manager_client.ts
```diff
@@ -254,6 +283,87 @@ export const ConnectionManagerClient = new class ConnectionManagerClient impleme
...
+
+	private async handlePasskeyLoginFinish({ credentialId, clientDataJSON, authenticatorData, signature }: IClientDirectoryArgument['passkeyLoginFinish'], connection: ClientConnection): IClientDirectoryPromiseResult['passkeyLoginFinish'] {
+		if (connection.isLoggedIn())
+			throw new BadMessageError();
+
+		const account = await accountManager.loadAccountByPasskeyCredentialId(credentialId);
+		if (!account || !account.secure.isActivated() || account.secure.isDisabled()) {
+			LoginManager.loginFailed();
+			return { result: 'unknownCredentials' };
+		}
```
Jomshir98: Disabled and non-activated accounts shouldn't produce unknown credential, but rather their respective results (verificationRequired / accountDisabled). Neither should cause the passkey to be deleted on client side for sure.

pandora-server-directory/src/networking/manager_client.ts
```diff
@@ -254,6 +283,87 @@ export const ConnectionManagerClient = new class ConnectionManagerClient impleme
...
+
+		const verification = this.verifyPasskeyAssertionForAccount(account, 'login', {
```
Jomshir98: It might be a good idea to verify the cryptography of the challenge answer before actually linking it to an account. That way we prove the user holds the passkey id, and we can avoid the constant time check for this endpoint - if you hold a valid passkey, you know the account exists in the first place. With that we can also separate response between verification failing, and the credential actually being unknown, as there is no guessable leak here.
We need to do that anyway, as verification can fail sporadically (e.g. timed out challenge or server reboot) and we certainly don't want to signal unknown passkey for those cases either.

pandora-server-directory/src/networking/manager_client.ts
```diff
@@ -1069,6 +1244,114 @@ export const ConnectionManagerClient = new class ConnectionManagerClient impleme
 		return { result: token == null ? 'invalidPassword' : 'ok' };
 	}
 
+	private handlePasskeyList(_: IClientDirectoryArgument['passkeyList'], connection: ClientConnection): IClientDirectoryResult['passkeyList'] {
+		if (!connection.isLoggedIn())
+			throw new BadMessageError();
+
+		return {
+			passkeys: connection.account.secure.listPasskeys(),
+			limit: LIMIT_ACCOUNT_PASSKEY_COUNT,
```
Jomshir98: Why is limit being returned as value, if it is a define in common? Also many places in client use that constant directly, so I see little point in this.

pandora-server-directory/src/networking/manager_client.ts
```diff
@@ -1069,6 +1244,114 @@ export const ConnectionManagerClient = new class ConnectionManagerClient impleme
...
+	private async handlePasskeyRegisterFinish({ name, credentialId, publicKey, clientDataJSON, authenticatorData, transports, cryptoKey }: IClientDirectoryArgument['passkeyRegisterFinish'], connection: ClientConnection): IClientDirectoryPromiseResult['passkeyRegisterFinish'] {
+		if (!connection.isLoggedIn())
+			throw new BadMessageError();
+
+		if (!connection.hasSudo())
+			return { result: 'sudoRequired' };
+
+		const existingCredentialAccount = await accountManager.loadAccountByPasskeyCredentialId(credentialId);
+		if (existingCredentialAccount != null && existingCredentialAccount !== connection.account)
+			return { result: 'alreadyExists' };
```
Jomshir98, line 1349: This is prone to race conditions. Though registering passkeys is a rare action, so maybe just globally synchronize the method for Pandora's Directory instance?

pandora-server-directory/src/account/accountPasskeys.ts
```diff
@@ -0,0 +1,465 @@
...
+			origin: parsed.origin,
+		};
+	} catch {
+		return null;
+	}
+}
+
+function ValidateRpIdHash(authenticatorData: Buffer): boolean {
+	const actual = authenticatorData.subarray(0, WEBAUTHN_AUTHENTICATOR_DATA_RP_ID_HASH_LENGTH);
+	const expected = createHash('sha256').update(PASSKEY_RP_ID, 'utf-8').digest();
```
Jomshir98: This hash could be cached

pandora-client-web/src/crypto/passkey.ts
```diff
@@ -0,0 +1,272 @@
...
+				name: 'Pandora',
+				id: start.rpId,
+			},
+			user: {
+				id: Base64UrlToArray(start.user.id),
+				name: start.user.name,
+				displayName: start.user.displayName,
+			},
+			pubKeyCredParams: [
+				{ type: 'public-key', alg: COSE_ALGORITHM_ES256 },
```
Jomshir98: We should probably allow more algorithms.
Yubico officially recommends at least RS256 and ES256 (https://developers.yubico.com/WebAuthn/WebAuthn_Developer_Guide/Algorithms.html).
It might also be a good idea to add Ed25519 for final list of [Ed25519, ES256, RS256] in this order.

pandora-server-directory/src/account/accountPasskeys.ts
```diff
@@ -0,0 +1,465 @@
...
+	]);
+	const key = createPublicKey({
+		key: Buffer.from(passkey.publicKey, 'base64'),
+		format: 'der',
+		type: 'spki',
+	});
+	if (!verify('sha256', signedData, key, Base64UrlDecode(data.signature)))
+		return { ok: false };
+	const signCount = authenticatorData.readUInt32BE(WEBAUTHN_AUTHENTICATOR_DATA_SIGN_COUNT_OFFSET);
+	if (passkey.signCount !== 0 && signCount !== 0 && signCount <= passkey.signCount)
```
Jomshir98: `signCount !== 0` must NOT be here. If the passkey returned sign count in the past, it MUST return sign count in the future, otherwise attacker cloning the authenticator could just always return zero to skip this check.

pandora-server-directory/src/account/accountPasskeys.ts
```diff
@@ -0,0 +1,465 @@
...
+		.replaceAll('+', '-')
+		.replaceAll('/', '_')
+		.replaceAll('=', '');
+}
+
+export function Base64UrlDecode(data: string): Buffer {
+	const normalized = data.replaceAll('-', '+').replaceAll('_', '/');
+	const padding = '='.repeat((4 - normalized.length % 4) % 4);
+	return Buffer.from(normalized + padding, 'base64');
+}
```
Jomshir98: Nitpick: Node's Buffer supports `base64url` directly.

Jomshir98 (approved):

PR #1135, opened 2026-06-26
[ADD] Several pose limits-related optimizations

Jomshir98:
## References

_None_

## Changelog

```md
Technical changes:
- Several performance optimizations related to pose limits:
  - Character restriction manager now evaluates character's properties lazily. This prevents several common cases where they were being calculated without being needed.
  - Check for being allowed to move a character in the room view is now more optimized.
  - Pose limit tree now attempts to combine pose tree children that are subset of one another. This prevents combinatorics from going rampart in the common case of holding many instances of an single asset with complex pose limits at the same time, holding more than 20 sharks should be possible now without Pandora blowing up... if you have that large arms.
```

## Checklist


ClaudiaMia (approved):
Amazing!

PR #1144, opened 2026-07-24
[FEATURE] API and CLI

Jomshir98:
## References

_None_

## About The Pull Request

This PR introduces initial architecture in preparation for BotAPI.
Specifically it creates new "pandora-api" library and "pandora-cli" tool with only example usage for now. Furthermore it introduces "advanced settings", which warn about why the settings are considered advanced and "(Personal) Access Tokens", which can be used in API to act as an account with specified scopes.

## Changelog

```md
Platform changes:
- Added a new "Advanced settings" category, which groups settings we consider advanced and shows appropriate warning and explanation when accessing them for the first time.
- Added "Access Tokens" menu, which can be used to give tools outside of Pandora limited access to your account without sharing your credentials. There are no useful tools yet - this is mainly a base on which more functionality can be developed.

Fixes:
- Fixed some text links not being usable with keyboard.

Technical changes:
- Introduced new pandora-api project. This project is intended to be used by external tools (e.g. in the future "Bots"). Currently it supports the following actions:
  - Connecting to the Directory server, automatic reconnection, and authentication using provided Personal Access Token.
  - Monitoring of account and token changes on Directory, so invalidation, regeneration, and token updates are propagated live to existing connections.
  - Specifications of "scopes" for the token, which limit what the API is allowed to do with it.
  - Several basic APIs:
    - Querying information about own token and invalidating it.
    - Space search for fully public spaces and active public spaces.
    - Listing owned spaces (requires `spaces:list_owned` scope)
    - Create new space (requires `spaces:create` scope)
    - Abandon an owned space (requires `spaces:disown` scope)
- Introduced new pandora-cli project providing simple commandline interface for using and testing some of API's features.
- Added helper types `Result` and `Option` that clarify API's usage.
- Updated build output version to ES2024.
- Small tooling and configuration cleanup.
```

## Checklist


pandora-server-directory/src/account/secure/accessTokens.ts
```diff
@@ -0,0 +1,204 @@
...
+			token,
+			id,
+			name,
+			scopes,
+			created: Date.now(),
+			expires,
+		};
+		this.#tokens.push(cloneDeep(tokenData));
+
+		await this.#accountSecure.updateDatabase();
```
Sekkmer, line 141: I see that we have a catch on this method above if it fails here or many other cases `#tokens` will different from db state as we exit

pandora-server-directory/src/account/secure/accessTokens.ts
```diff
@@ -0,0 +1,204 @@
...
+			throw new Error('Encountered a unicorn! (generated duplicate access token)');
+		}
+
+		let id: string;
+		do {
+			id = nanoid(8);
+		} while (this.#tokens.some((t) => t.id === id));
+
+		const tokenData: PandoraAccessTokenData = {
+			token,
```
Sekkmer: maybe we could store SHA-256 or a HMAC-SHA-256 so only user have the real token we only need to validate this

pandora-common/src/account/accessTokens.ts
```diff
@@ -0,0 +1,184 @@
...
+	readonly description: string;
+	/** If this scope requires other scopes, NOT transitive */
+	readonly requires?: readonly PandoraAccessTokenScope[];
+};
+
+export type PandoraAccessTokenScope = keyof typeof PANDORA_ACCESS_TOKEN_SCOPES_DEFINITION;
+export const PandoraAccessTokenScopeSchema: z.ZodType<PandoraAccessTokenScope> = z.enum(KnownObject.keys(PANDORA_ACCESS_TOKEN_SCOPES_DEFINITION));
+
+export type PandoraAccessTokenScopeList = readonly PandoraAccessTokenScope[];
+export const PandoraAccessTokenScopeListSchema: z.ZodType<PandoraAccessTokenScopeList> = PandoraAccessTokenScopeSchema.array()
```
Sekkmer, line 43: hmm maybe require this to be unique, and how about enforcing require?

pandora-server-directory/src/account/secure/accessTokens.ts
```diff
@@ -0,0 +1,204 @@
...
+			created: t.created,
+			lastUsed: t.lastUsed,
+			expires: t.expires,
+		}));
+	}
+
+	private readonly _delayedUpdateDebounced = debounce(() => {
+		this.#accountSecure.updateDatabase()
+			.catch((err) => {
+				this._auditLog.warning('Failed to update lastUsed on token:', err);
```
Sekkmer, line 76: how about logging creation, deletion, regeneration, expiration changes, and scope changes too?

pandora-server-directory/src/account/accountManager.ts
```diff
@@ -243,6 +243,20 @@ export class AccountManager implements ServerService {
 		return null;
 	}
 
+	/**
+	 * Find an account between **currently loaded accounts**
+	 * @returns The account or `null` if not found
+	 */
+	public getAccountByAccessToken(token: PandoraAccessToken): Account | null {
+		for (const account of this._onlineAccounts) {
+			if (account.secure.accessTokens.verifyToken(token, EMPTY_ARRAY)) {
```
Sekkmer: might not matter but long run might worth adding a lookup, though with how many users we have now it should not matter

pandora-server-directory/src/networking/api/socket/connection_api.ts
```diff
@@ -0,0 +1,104 @@
...
+	public verifyTokenUse(requiredScopes: readonly PandoraAccessTokenScope[]): boolean {
+		if (this._account == null)
+			return false;
+
+		if (this._account.secure.accessTokens.verifyToken(this.token, requiredScopes)) {
+			return true;
+		}
+
+		// Check if the token was altogether invalidated
+		if (!this._account.secure.accessTokens.verifyToken(this.token, EMPTY_ARRAY)) {
```
Sekkmer: maybe we could have a verifyTokenDetaled with valid missing scope, invalid/expired

pandora-server-directory/src/account/secure/accessTokens.ts
```diff
@@ -0,0 +1,204 @@
...
+
+const AccessTokenSecretGenerator = nanoCustomAlphabet('0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ');
+
+const tokenUseMetric = new promClient.Counter({
+	name: 'pandora_directory_access_token_usage',
+	help: 'Count of uses of access tokens by scope. Use requiring multiple scopes counts multiple times.',
+	labelNames: ['scope'],
+});
+
+const LAST_USE_UPDATE_DEBOUNCE = 5000; // Update "last use" value only at most every 5 seconds
```
Sekkmer, line 32: why do we need to update database this fast, not sure how frequent the token is validated though, if a bot use a token too much could add some charn

pandora-api/src/internal/socketio_connector.ts
```diff
@@ -0,0 +1,92 @@
...
+	messageHandler: IMessageHandler<IncomingT>;
+	onConnect: () => void;
+	onDisconnect: (reason: Socket.DisconnectReason) => void;
+	onConnectError: (err: Error) => void;
+	logger: Logger;
+}
+
+type SocketAuthCallback = (data?: object) => void;
+
+export const URL_PATH_REGEX = /^((?:(https?|wss?):\/)?\/?([^:/\s]+)(?::(\d+))?)(\/[^#\s]+)?$/;
```
Sekkmer: wouldn't `new URL()` be simpler and more robust?

pandora-client-web/src/components/settings/personalAccessTokensSettings/tokenList.tsx
```diff
@@ -0,0 +1,122 @@
...
+				AssertNever(result);
+			}
+		},
+	);
+
+	return (
+		<tr>
+			<td>
+				<strong>
+					<a onClick={ (ev) => {
```
Sekkmer: without href this is might not be keyboard selectable

pandora-client-web/src/components/settings/personalAccessTokensSettings/tokenList.tsx
```diff
@@ -0,0 +1,122 @@
...
+					new Date(token.lastUsed).toLocaleString()
+				) }
+			</td>
+			<td>
+				<ButtonConfirm
+					theme='danger'
+					slim
+					disabled={ !sudoActive || processing }
+					title='Delete token'
+					content={ `Are you sure you want to delete token ${token.name}?\nAny application using this token will loose access to your account.` }
```
Sekkmer: typo: lose

pandora-api/src/internal/socketio_connector.ts
EllInRepose, line 1: This seems like a duplicate of pandora-client-web\src\networking\socketio_connector.ts that would drift and need patching independently, move to common and share the functionality, deduplicate code perhaps?

pandora-server-directory/src/account/secure/accessTokens.ts
```diff
@@ -76,7 +76,7 @@ export class AccountSecureAccessTokenStore extends TypedEventEmitter<{
 		const tokenData = this.#tokens.find((t) => t.token === token);
 
 		const now = Date.now();
-		if (tokenData == null || (tokenData.expires != null && now < tokenData.expires))
+		if (tokenData == null || (tokenData.expires != null && now >= tokenData.expires))
```
EllInRepose, line 91: Ooph nice catch, though I would expect some tests to change too with some very critical functionality like this

EllInRepose:
/pandora-api/node-modules/ definitely need to be added to the .gitignore. Ask me how I know

pandora-server-directory/src/account/secure/accessTokens.ts
Jomshir98: This is problem of two generals (if DB request succeeds but its confirmation is not returned successfully, then proposed change has identical problem). We can either move all requests to the DB directly, or just leave it be as we never had a DB failure as far as I know.
I would be for keeping it like this.

pandora-common/src/account/accessTokens.ts
Jomshir98: It already does `.max(KnownObject.keys(PANDORA_ACCESS_TOKEN_SCOPES_DEFINITION).length)` so it can't be longer than list of all unique scopes (we use this simple check in several places that accept a set-like argument already).
As for enforcing require, I'm not sure that should be done at the types level - requirements can change over time, so how would that affect existing tokens in DB? Should it drop them? Add new scopes to them without user prompt? Invalidate existing scopes?

pandora-server-directory/src/account/accountManager.ts
Jomshir98: I think not worth it now, as only time this happens is on a _new_ API socket connection (not on each request) and keeping up-to-date map would be rather annoying, as tokens can change fair bit in theory.

pandora-server-directory/src/account/secure/accessTokens.ts
Jomshir98: I think this amount of updating is fine. Note that any increase also technically increases Directory shutdown time. The update is generally quite cheap, so this update rate is just a de-bounce for really rapidly used tokens. Also I expect common use-case to be rare, short burst of several requests, which this covers.

pandora-api/src/internal/socketio_connector.ts
Jomshir98: pandora-api is intentionally avoiding usage of platform-specific features and `URL` is not a core ECMAScript feature. `new URL()` is available in both browser and Node, but there are no "browser or node" typings to use here.

pandora-api/src/internal/socketio_connector.ts
Jomshir98: Connectors are not really standardized right now and this slight difference is intentional - they do need _slightly_ different approach. Refactoring this (and Client's Shard connector and Shard's Directory connector) is planned, but rather low priority.
I'm honestly hoping that WebTransport gets standardized soon-ish so we can just use that, but yeh, not holding my breath...

pandora-server-directory/src/account/secure/accessTokens.ts
Jomshir98: But buuut... who wants to write those! Besides I do a lot of manual testing, which is how this got caught! *nods nods*

Jomshir98:
> /pandora-api/node-modules/ definitely need to be added to the .gitignore. Ask me how I know

We do per-project gitignores - `node_modules` are correctly ignored in both `pandora-api` and `pandora-cli` projects, but the projects are new - so if you switch between commits that have and don't have them, you will have to clean up manually.

pandora-server-directory/src/database/mongoDb.ts
```diff
@@ -170,6 +170,13 @@ const accountCollection = new ValidatedCollection(
 			// Ignore accounts without passkeys
 			sparse: true,
 		},
+		{
+			name: 'accountAccessToken',
+			unique: true,
+			key: { 'secure.accessTokens.token': 1 },
```
Sekkmer: shouldn't this be changed to `tokenHash` too

pandora-server-directory/src/networking/api/socket/connection_api.ts
```diff
@@ -0,0 +1,110 @@
...
+
+		const verifyResult = this._account.secure.accessTokens.verifyToken(this.tokenHash, requiredScopes);
+		if (verifyResult === 'ok') {
+			return true;
+		}
+
+		// Check if the token was altogether invalidated and disconnect the connection if yes
+		if (verifyResult === 'disabledAccount' || verifyResult === 'invalidToken') {
+			queueMicrotask(() => {
+				this.disconnect('token expired');
```
Sekkmer, line 102: hmm is disabled account really an expired token? could give more detailed issue here maybe

pandora-client-web/src/components/common/link/interactiveLink.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export interface InteractiveLinkProps extends CommonProps, Pick<ComponentPropsWithRef<'a'>, 'ref' | 'title'> {
+	onClick: () => void;
+}
+
+/**
+ * A link (`<a>`) element that does not perform navigation, but triggers a code callback instead.
+ * Supports both mouse and keyboard triggers.
+ */
+export function InteractiveLink({ onClick, children, ref, ...props }: InteractiveLinkProps): ReactElement {
```
Sekkmer, line 12: think the previous iteration had `stopPropagation()` not sure but might be needed here as well

pandora-api/src/internal/socketio_connector.ts
Sekkmer: I agree that importing Node- or browser-specific URL implementations would undermine the platform-neutral design. However, `globalThis.URL` is provided by both intended Node and browser runtimes. The current regex has several concrete parsing differences and its return type is not always accurate.

- `http://[::1]:25560/api_socket.io` — valid IPv6 URL, but rejected.
- `HTTPS://example.com/api_socket.io` — valid case-insensitive scheme, but rejected.
- `https://example.com:99999/api_socket.io` — invalid port, but accepted.
- `https://example.com?token=test` — query is incorrectly included in `origin`.
- `https://example.com#fragment` — fragment is incorrectly included in `origin`.
- `https://example.com/api_socket.io?token=test` — query is incorrectly included in `path`.
- `https://example.com` — returns `path: undefined` despite declaring `path: string`.

Use the standard global URL implementation through a narrow platform-neutral structural type:

```ts
interface UniversalURL {
	readonly origin: string;
	readonly pathname: string;
	readonly protocol: string;
	readonly username: string;
	readonly password: string;
	readonly search: string;
	readonly hash: string;
}

type UniversalURLConstructor = new (input: string) => UniversalURL;

const URLConstructor = (
	globalThis as typeof globalThis & {
		URL: UniversalURLConstructor;
	}
).URL;

const url = new URLConstructor(uri);

if (!['http:', 'https:', 'ws:', 'wss:'].includes(url.protocol) ||
	url.username !== '' ||
	url.password !== '' ||
	url.search !== '' ||
	url.hash !== '') {
	throw new Error(`Invalid Socket.IO URL ${uri}`);
}

return {
	origin: url.origin,
	path: url.pathname,
};
```

This remains runtime-portable between Node and browsers without enabling all DOM typings or maintaining a custom URL parser.

pandora-common/src/account/accessTokens.ts
Sekkmer: unsure maybe not at the type level though

pandora-api/src/internal/socketio_connector.ts
Jomshir98: The only valid case here from all the urls is the IPv6. If I'm really generous then the case-insensitive scheme too, but that can be fixed by just adding `i` flag.
All others are either user error or can't even theoretically result in successful connection, as path is required. And worst-case impact is the api/cli failing to connect for the the very user that entered wrong _manual_ value (which is an advanced feature in the first place, with named values provided).

Nevertheless, updated.

pandora-server-directory/src/database/mongoDb.ts
Jomshir98: Good catch! _I hate how weakly-typed those keys are_

pandora-client-web/src/components/common/link/interactiveLink.tsx
Jomshir98: Intentionally removed. The more I learn about good practices, the more I have tendency to rewrite old things... but not worth looking for such small mistakes everywhere.
These days I would rather check "default prevented" on parent than stop propagation, if there would be two nested interactive elements... I don't think there are any link-based ones left though.

pandora-server-directory/src/account/secure/accessTokens.ts
Jomshir98: Here, does this look better?
image
(the "functions" coverage is a coverage bug related to decorators)

EllInRepose: <3

pandora-cli/src/cliCommandRunner.ts
```diff
@@ -0,0 +1,45 @@
...
+	if (commandArgs.length === 0 || !commandArgs[0].trim()) {
+		process.stdout.write(GetHelp());
+		return;
+	}
+
+	const logger = GetLogger('Command');
+
+	const command = CLI_COMMANDS.find((c) => c.key === commandArgs[0]) ?? null;
+	if (!command) {
+		logger.fatal(`Unknown command "${commandArgs[0]}"`);
```
Sekkmer, line 17: `process.exitCode = 1`

pandora-cli/src/commands/spaceSearch.ts
```diff
@@ -0,0 +1,158 @@
...
+
+							if (format === 'text') {
+								if (result.value.length === 0) {
+									process.stdout.write('No spaces found\n');
+								} else {
+									for (const space of result.value) {
+										process.stdout.write(
+											`- Id: ${space.id}\n` +
+											`  Name: ${space.name}\n` +
+											`  Description: ${space.description.includes('\n') ? ('|\n    ' + space.description.replaceAll('\n', '\n    ')) : space.description}\n` +
```
Sekkmer, line 36: space name are unsanitized though issue is minor one could set a public space name `"Welcome to the museum\u001b[2JFAKE CLI OUTPUT"`

pandora-api/src/api/apis/spaceManagement.ts
```diff
@@ -0,0 +1,113 @@
...
+	 * @returns Result containing either id of the new space or reason for failure.
+	 */
+	public async createSpace(config: SpaceDirectoryConfig): Promise<Result<{
+		id: SpaceId;
+	}, CreateSpaceError>> {
+		const parsedConfig = SpaceDirectoryConfigSchema.safeParse(config);
+		if (!parsedConfig.success) {
+			return Result.Err({
+				type: 'invalidArgument',
+				parseError: parsedConfig.data,
```
Sekkmer: shouldnt this be: `parsedConfig.error`?

pandora-server-directory/src/networking/api/socket/handlers/spaceManagement.ts
```diff
@@ -0,0 +1,59 @@
...
+			return { result: 'notAllowed' };
+
+		const space = await SpaceManager.loadSpace(spaceId);
+
+		if (space == null) {
+			GetLogger('ApiHandlersSpaceManagement').verbose(`${connection.id} failed manage space ownership: Space not found`);
+			return { result: 'notFound' };
+		}
+
+		const result = await space.removeOwner(account.id);
```
Sekkmer, line 45: hmm this will delete the space on last owner unsure maybe a `space:delete` extra token role could be added for that, or maybe some form of `dryRun: boolean`

pandora-server-directory/src/networking/api/socket/handlers/spaceManagement.ts
```diff
@@ -0,0 +1,59 @@
...
+
+/** API message handlers related to space management */
+export const ApiHandlersSpaceManagement = {
+	spaceCreate: async ({ config }, connection): IApiDirectoryPromiseResult['spaceCreate'] => {
+		const account = connection.verifyTokenUseAndGetAccount(['spaces:create']);
+		if (account == null)
+			return { result: 'notAllowed' };
+
+		// Only developers can create rooms with development mode enabled
+		if (config.features.includes('development') && !account.roles.isAuthorized('developer')) {
```
Sekkmer: is it a good idea to duplicate this check? this could be uniform in any space creation API or WebUI

pandora-server-directory/src/networking/api/socket/handlers/spaceManagement.ts
Jomshir98: This is how disowning always worked. From the (past) owner's point of view this is not any different really - they loose access to it either way, no matter if someone else retains it or not.
Client also does not differentiate between this in any way.

And what would be the point of dry-run? What would it return as result, if you can already do `spaceOwnedList` and will be able to get more info about the space in the future, if assigned to it.

pandora-api/src/api/apis/spaceManagement.ts
Jomshir98: Nice catch!

pandora-cli/src/commands/spaceSearch.ts
Jomshir98: Space names are checked:
```ts
export const LIMIT_SPACE_NAME_PATTERN = /^[a-zA-Z0-9_+\-:'"& ]+$/;
```
As for descriptions... honestly, it would be better to forbid control/non-printable characters than trying to sanitize it here. Also not sure what anyone would gain by doing that? I don't expect CLI to actually be used in practice - API would be much more likely used, and then it is up to each dev to sanitize what they print.
Also I don't believe that injecting output to console carries any significant risk.

Sekkmer (approved):

